PR #57, opened 2017-08-12
fix link

cherti:
move pentabarf-link to the github-page as the page http://pentabarf.org seems to be dead.


rixx:
Thanks! :smile:

PR #61, opened 2017-08-20
Allow talk lookup via talk PK

andir:


rixx:
Fixed it via afe4fec instead, as communicated elsewhere :)

PR #63, opened 2017-08-23
add entries from github python.gitignore to .gitignore

luto:
This adds entries from [github python.gitignore](https://github.com/github/gitignore/blob/master/Python.gitignore) to `.gitignore`. It helps new developers, who might use a different tool chain to get started with this project.


rixx:
Thanks a lot!

PR #71, opened 2017-08-24
[orga] try to make the buttons in schedule editor clearer

luto:
The schedule editor already looks great, but the buttons on the top kind of confused me. I tried to give them better names. Please note that this is only a suggestion, feel free to make or request any changes you want :)


rixx:
Thank you very much! UI testers who write their own PRs are the best UI testers. :bee:

PR #74, opened 2017-08-25
[agenda] ensure that talks for non-public events are inaccessible via their url

luto:
Events that are not public yet just display a 404-page for their schedule and main public pages. Talks should be hidden as well, just in case the url is shared somewhere by accident.

### DIFF
--- a/src/pretalx/agenda/views/schedule.py
+++ b/src/pretalx/agenda/views/schedule.py
@@ -160,7 +160,7 @@ class FrabJsonView(ScheduleDataView):
         return JsonResponse({'schedule': result})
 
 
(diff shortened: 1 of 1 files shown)


rixx:
This just shows how very much `EventPageMixin` needs to be a middleware. Thanks!

raphaelm:
There's more to it here. You just need *one* public event to access *all* submissions, regardless of their state or which event they belong to, through this view. We'd need a ``get_queryset`` method in that view.

luto:
@raphaelm good catch, thanks! I wrote up a quick issue so this doesn't get lost.

PR #75, opened 2017-08-25
[agenda] make speaker links on public talk pages bold, as planned

luto:
The agenda scss contains a selector `a.speaker` to make speaker links on public talk pages bold. But there is no `<a class="speaker">` on that page. Only a `<div class="speaker"><a>`. This PR reflects the actual HTML structure in CSS.


luto:
oh. nvm. m)

PR #76, opened 2017-08-25
[agenda] add speaker page

### LINKED ISSUE #53: Link all talks of a speaker on every of their talks

luto:
Add a speaker detail page for the public agenda. The new page displays the speakers nickname, name as well as a list of talks the speaker holds. There is also a link to the speaker page on each of their talks.

I kind of took the idea of having a public speaker page and ran with it. In case you do not like it or you want something changed, just tell me so and I'll adapt this PR.

kind-of-fixes #53?

### DIFF
--- a/src/pretalx/agenda/templates/agenda/speaker.html
+++ b/src/pretalx/agenda/templates/agenda/speaker.html
@@ -1,0 +1,38 @@
+{% extends "agenda/base.html" %}
+{% load i18n %}
+
+{% block content %}
+<h3>
+    {{ speaker.nick }}
+    {% if speaker.name %}
+        <small>{{ speaker.name }}</small>
+    {% endif %}
+</h3>
+
+<div class="row">
+    <div class="col-lg-8 col-md-6 col-xs-12">
+        {% if speaker.biography %}
+            {{ speaker.biography }}
+        {% else %}
+            {% trans 'This speaker did not provide a biography yet.' %}
+        {% endif %}
+    </div>
+    <div class="col-lg-4 col-md-6 col-xs-12">
+        <div class="bordered-list">
+            <div class="header">
+                {% blocktrans trimmed count count=submissions.count %}
+                Talk
+                {% plural %}
+                Talks
+                {% endblocktrans %}
+            </div>
+            {% for submission in submissions %}
+                <div class="info">
+                    <div><a href="{{ submission.urls.public }}">{{ submission.title }}</a></div>
+                    <div>{{ submission.description }}</div>
+                </div>
+            {% endfor %}
+        </div>
+    </div>
+</div>
+{% endblock %}
--- a/src/pretalx/static/agenda/scss/_speaker.scss
+++ b/src/pretalx/static/agenda/scss/_speaker.scss
@@ -18,30 +18,31 @@
       border-bottom: 1px solid rgba(0, 0, 0, 0.1);
     }
   }
+}
 
-  .speakers {
-    border: 1px solid #4CAF50;
-    border-top: 4px solid #4CAF50;
-    background-color: rgba(255, 255, 255, 0.76);
-    box-sizing: border-box;
-    color: rgba(0, 0, 0, 0.87);
-    display: block;
-    padding: 5px 10px;
+/* used for speaker list on talk page as well as talk list on speaker page */
+.bordered-list {
+  border: 1px solid #4CAF50;
+  border-top: 4px solid #4CAF50;
+  background-color: rgba(255, 255, 255, 0.76);
+  box-sizing: border-box;
+  color: rgba(0, 0, 0, 0.87);
+  display: block;
+  padding: 5px 10px;
 
-    .speaker-header {
-      font-weight: bold;
-    }
+  .header {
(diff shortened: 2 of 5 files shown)


luto:
image

Looks like this :) If there is a "real name" it's displayed in grey beneath the nickname.

src/pretalx/agenda/templates/agenda/speaker.html
```diff
@@ -0,0 +1,38 @@
...
+        <small>{{ speaker.name }}</small>
+    {% endif %}
+</h3>
+
+<div class="row">
+    <div class="col-lg-8 col-md-6 col-xs-12">
+        {% if speaker.biography %}
+            {{ speaker.biography }}
+        {% else %}
+            {% trans 'This speaker did not provide a biography yet.' %}
```
luto: This could probably also say "$SpeakersNickname (or $SpeakersName?) did not provide a biography yet.", which sounds a little nicer. What do you think?

src/pretalx/static/agenda/scss/_speaker.scss
```diff
@@ -18,30 +18,31 @@
...
-  .speakers {
-    border: 1px solid #4CAF50;
-    border-top: 4px solid #4CAF50;
-    background-color: rgba(255, 255, 255, 0.76);
-    box-sizing: border-box;
-    color: rgba(0, 0, 0, 0.87);
-    display: block;
-    padding: 5px 10px;
+/* used for speaker list on talk page as well as talk list on speaker page */
+.bordered-list {
```
luto: In case you come up with a nicer name for this class, let me know. Maybe I can think of one tomorrow.

rixx:
This is a tricky one. The issue #53 was phrased that way for a reason – we actually have two issues here that are equally hard to solve.

Speakers may not want their nicks (which are login relevant user names) to be public, which is to my mind a very reasonable assumption. This leaves us without a great way to choose a URL for a speaker page, since the only other identifying property is the ID, and I really do not want to expose IDs.

The second issue is that in terms of usability, I don't believe a speaker page to be all that useful. It mostly adds overhead – I'll rarley ever look for a speaker's page, I'll look for a talk, and then look for other talks by that speaker from there. So putting this information in one place makes sense (to me, at least).

frab has a speaker list view, and I guess I could be persuaded to put something like this in, but I'd rather avoid it.

I know this sucks, since you already put the work in this PR – my mind is not completely set here, so if you think I'm wrong, go ahead and tell me ;)

luto:
> Speakers may not want their nicks (which are login relevant user names) to be public, which is to my mind a very reasonable assumption.

The `name`-field is not required. If the nick is no good for displaying speakers publicly and the name is both not required and not unique, which field should be used in the UI? For example in the list of speakers on the talk page or the not-yet-existing big speaker list. Use the name, if given, otherwise fall back to the nickname? Use the name, otherwise fall back to "Anonymous"?

> (... ) a speaker page (...) mostly adds overhead – I'll rarley ever look for a speaker's page, I'll look for a talk, and then look for other talks by that speaker from there. So putting this information in one place makes sense (to me, at least).

If there is only one speaker, it's rather easy, I think. Just add a list beneath the page, which will almost always contain zero, and sometimes 2 entries.

For me it gets complicated once there is more than one speaker. In a realistic worst-case, a talk would have 3 speakers, each of which has one other talk so we'd need to display 6 - each with some kind of hopefully short description. Some of these talks may be shared across some of those speakers. I am not sure how to indicate which talk belongs to which speaker here: Just display a table, which has a list of speakers for each talk in one column? Display a list of talks for each speaker, possible causing duplicates on that page? Display a list of all talks, ignoring which of the speakers actually holds them?

Also, I quickly see the "other talks from these speakers"-section becoming bigger than the "information about the talk"-section on the talk pages, which worries me a bit. Right now the page is just so neat and tidy.

But maybe I'm making too much of fuss here. What do you think?

> I know this sucks, since you already put the work in this PR

ha, if I want PRs that go through every time, I need to talk about the code I'm going to write beforehand, not afterwards ;) it's all good. If everything this PR does is to provide a prototype, so we or you can decide on how to display speaker information, it has done its job.

rixx:
> The name-field is not required. If the nick is no good for displaying speakers publicly and the name is both not required and not unique, which field should be used in the UI?

We're talking about two different things here – in *display,* the nickname is currently displayed when no name is given. I'd say this is not very well thought through, but it works for now, although we should consider making the name field mandatory and calling it "Display name" in the UI. I see this as less critical because the speaker can change this by setting her name.

URLs are meant to be stable, so the non changeable nickname would be the best solution … I just don't like that it is also a log in credential. :\
We could, of course, require the email for log in and use the nickname as a slug … Does that sound better?

I agree with your UI concerns – but they may be unfounded. If you just provide a "Other talks" list for the speaker, even three speakers with two talks apiece would only in three additional bullet points with a talk title, which is not too much, I think?

Regardless of if we build a speaker page now, we need this problem solved. I'd imagine that global speaker pages would be quite cool ('She held this talk at MRMCD 2017 and that workshop at MRMCD 2018.').

luto:
Abandoned in favor of #87.

PR #77, opened 2017-08-25
[orga] display a friendy message when the user deletes a room, which is in use

### LINKED ISSUE #73: deleting a room, which is in use yields ProtectedError
When you try to delete a room, which has talks assigned to it the app displays an ProtectedError. This can probably be handled it [deleting a submission type](https://github.com/openeventstack/pretalx/blob/master/src/pretalx/orga/views/cfp.py#L147-L152).

luto:
Fixes #73.

Please refer to the issue for details :)

### DIFF
--- a/src/pretalx/orga/views/settings.py
+++ b/src/pretalx/orga/views/settings.py
@@ -3,6 +3,7 @@ import string
 from django.contrib import messages
 from django.contrib.auth import login
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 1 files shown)


rixx:
Seeing as the pretalx code is ridiculously prone to errors such as this at the moment, I'm very grateful that you take the time to submit PRs like this :)

luto:
Sure thing :) clicking around to make things break (and then actually fix them) is fun :D

PR #83, opened 2017-08-27
[schedule] make sure the wip_schedule cache is refreshed on freeze()

luto:
fixes the following bug:
```python
>>> event.wip_schedule
<Schedule: None>
>>> event.wip_schedule.freeze('Version 2')
>>> event.wip_schedule
<Schedule: Version 2>
(expected: <Schedule: None>)
```

This should, in theory, also happen with `unfreeze()`, but I am not able to replicated it.


rixx:
Thanks! :cookie:

PR #84, opened 2017-08-27
Fix #72: schedule reset hides talk

### LINKED ISSUE #72: resetting schedule to an old version can make submissions unavailable
1. create submission: sub1
2. publish a schedule containing only sub1: ScheduleA
3. create submission: sub2
4. publish a schedule containing both sub1 and sub2: ScheduleB
5. reset the current draft to ScheduleA

Sub2 is not part of the current draft and should therefore be visible in the list on the right. But the list is empty. Resetting the draft to ScheduleB restores the talk.

luto:
This fixes #72. Please refer to the issue for details. It also adds a test or two for `Schedule.unfreeze()`.

### DIFF
--- a/src/pretalx/schedule/models/schedule.py
+++ b/src/pretalx/schedule/models/schedule.py
@@ -69,13 +69,25 @@ class Schedule(LogMixin, models.Model):
         from pretalx.schedule.models import TalkSlot
         if not self.version:
             raise Exception('Cannot unfreeze schedule version: not released yet.')
+
+        # collect all talks, which have been added since this schedule (#72)
+        submission_ids = self.talks.all().values_list('submission_id', flat=True)
+        talks = self.event.wip_schedule.talks \
+            .exclude(submission_id__in=submission_ids) \
+            .union(self.talks.all())
+
+        wip_schedule = Schedule.objects.create(event=self.event)
+        new_talks = []
+        for talk in talks:
+            new_talks.append(talk.copy_to_schedule(wip_schedule, save=False))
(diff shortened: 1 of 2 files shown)


src/pretalx/schedule/models/schedule.py
```diff
@@ -69,13 +70,25 @@ def unfreeze(self, user=None):
         from pretalx.schedule.models import TalkSlot
         if not self.version:
             raise Exception('Cannot unfreeze schedule version: not released yet.')
+
+        submission_ids = [talk.submission_id for talk in self.talks.all()]
+        new_talks = [
```
rixx: Consider going with

```python
new talks = self.event.wip_schedule.talks.exclude(talk__submission_id__in=self.talks.all().values_list('submission_id', flat=True)
```

src/pretalx/schedule/models/schedule.py
```diff
@@ -69,13 +70,25 @@ def unfreeze(self, user=None):
...
+            if t.submission_id not in submission_ids
+        ]
+
         self.event.wip_schedule.talks.all().delete()
         self.event.wip_schedule.delete()
+
         wip_schedule = Schedule.objects.create(event=self.event)
         talks = []
-        for talk in self.talks.all():
+        for talk in itertools.chain(new_talks, self.talks.all()):
```
rixx: Alternatively, if you use a queryset instead of a list in L75, you could go with the [union](https://docs.djangoproject.com/en/dev/ref/models/querysets/#union) method of Django querysets, I think?

rixx:
Thanks – looks great :)

PR #89, opened 2017-08-27
fix #86 add speaker textbox width

### LINKED ISSUE #86: orga new speakers textbox is too wide
image

The orga speakers page of a submission (`/orga/event/foo/submissions/1/speakers`) has a textbox to add new speakers. It works great, but looks a little wonky, because it takes up all the vertical space. This causes the button to be beneath the box, instead of next to it.

luto:
This takes the existing layout and iconography  of the team to style the talk speakers list.
image

Fixes #86.


rixx:
Woo-hoo! Thanks :)

PR #95, opened 2017-08-28
[orga] make the "New Event" button on the orga frontpage more consitent

luto:
image

The next inside the "New Event" button is all in one line, which can look a little wonky sometimes. Breaking it into two lines looks nicer, in my opinion:
image

### DIFF
--- a/src/pretalx/locale/de/LC_MESSAGES/django.po
+++ b/src/pretalx/locale/de/LC_MESSAGES/django.po
@@ -1258,8 +1258,8 @@ msgid "Dates and location"
 msgstr "Datum und Ort"
 
 #: pretalx/orga/templates/orga/dashboard.html:23
-msgid "Uh, a new event? Head over here, please!"
-msgstr "Ui, eine neue Veranstaltung? Hier entlang, bitte!"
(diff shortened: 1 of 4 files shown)


rixx:
Hahah, would you mind shortening the test to "Uh, a new event?"? – It's just to test that we are a) logged in and b) redirected to the cross-event dashboard.

luto:
Welp. 😁 Fixed.

rixx:
Hmm, another consistent solution would be to define a maximum width for dashboard tiles and letting normal text rendering take care of the line break. What do you think?

luto:
Letting normal text rendering do the job means that the "Uh, a new event?"-text will again break at some random point. There seems to be a way around that [by wrapping the lines into `<span>`s](https://stackoverflow.com/a/24357132/2486196), but it doesn't seem to be a clean solution in for our use case.

How about we define a maximum width *and* handle the "Uh, a new event?"-text via `<br>`, like before. I adapted the code to do this.

image

rixx:
I'll merge it and I suppose we will revisit it when people have more than a few tiles. :) Thanks!

PR #100, opened 2017-08-29
Fix frab import

luto:
Events created by frab import do not have a mail address. This makes the import crash. The speaker mail addresses are also set the "name@localhost". Sending mails is therefore pretty useless right now anyway. Even if there were proper email addresses, I'm not so sure about generating mails for imported talks.

This just disables sending emails during import. It also adds a very simple frab import test.

```
  File "/home/luto/c3w/pretalx/venv/lib/python3.6/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "/home/luto/c3w/pretalx/venv/lib/python3.6/site-packages/django/db/backends/sqlite3/base.py", line 328, in execute
    return Database.Cursor.execute(self, query, params)
django.db.utils.IntegrityError: NOT NULL constraint failed: mail_queuedmail.reply_to
```

### DIFF
--- a/src/pretalx/orga/management/commands/import_frab.py
+++ b/src/pretalx/orga/management/commands/import_frab.py
@@ -83,5 +83,5 @@ class Command(BaseCommand):
                     )
 
         schedule_version = root.find('version').text
-        event.wip_schedule.freeze(schedule_version)
+        event.wip_schedule.freeze(schedule_version, notify_speakers=False)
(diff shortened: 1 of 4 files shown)


rixx:
Uh, nice! :pizza:

luto:
🍕👍 😁

Issue #113, opened 2017-09-04
Fix event slug URL matching


rixx:
Event slugs in URLs match `\w`, but `-` is also allowed in event slugs, leading to inaccessible events.

PR #116, opened 2017-09-06
fix event slug url matching

luto:
See #113 for details :)

Django does not seem to provide an easy way to match slugs within URLs. [The internet](https://stackoverflow.com/questions/31216363/url-pattern-for-alphanumericslug-string-in-django) [recommends](https://stackoverflow.com/questions/23358685/regular-expression-in-url-for-django-slug) to just use an regex for that. So I did. If you know a smarter way, just hit me up and I'll change this around.

The diff is kind of big, because I removed all the duplicate slug-matching from the URLs. This way I only had the replace the regex in three places.

### DIFF
--- a/src/pretalx/agenda/urls.py
+++ b/src/pretalx/agenda/urls.py
@@ -1,18 +1,20 @@
-from django.conf.urls import url
+from django.conf.urls import include, url
 
 from .views import feed, location, schedule, speaker
 
 agenda_urls = [
-    url('^(?P<event>\w+)/schedule/$', schedule.ScheduleView.as_view(), name='schedule'),
-    url('^(?P<event>\w+)/schedule/changelog$', schedule.ChangelogView.as_view(), name='schedule.changelog'),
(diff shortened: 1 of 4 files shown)


luto:
> Imports are incorrectly sorted.

I really need a commit hook for that.

luto:
The failed pylama tests seem to be in `setup.py`, which is not touched by this PR.

rixx:
Hmm, this PR allows dashes, but does not include dots.

How about adding a `SLUG_CHARS = 'a-zA-Z0-9.-'` to `events.models.event`, and using it both in the `urls.py`s and in the `RegexValidator` on an event's slug field?

luto:
Sure, will do. :)

luto:
@rixx please take another look :)

rixx:
Thank you!

luto:
42 open issues in the tracker, 42 open issues.
Assign one to me and merge it around, 41 open issues in the tracker! 🎶

Issue #122, opened 2017-09-10
add <meta> tags for popular social networks


luto:
Facebook and twitter parse special `<meta>` tags on linked pages, to include some of the content right into their interface. While both [twitter](https://dev.twitter.com/cards/getting-started) and [facebook](https://davidwalsh.name/facebook-meta-tags) specify their own tags, twitter seems to parse both variants. So the minimal useful set of tags - according to twitter - is:
```html
<meta name="twitter:card" content="summary" />
<meta name="twitter:site" content="@twitternick" />
<meta property="og:url" content="http://absolute/url" />
<meta property="og:title" content="A Twitter for My Sister" />
<meta property="og:description" content="In the early days, Twitter grew so quickly." />
<meta property="og:image" content="http://graphics8.nytimes.com/images/2011/12/08/technology/bits-newtwitter/bits-newtwitter-tmagArticle.jpg" />
```
Most of that information is already present. The twitter nick could maybe be a new conference setting? It can also be omitted for now.

luto:
add ALL the meta tags! 🎉 😁

PR #123, opened 2017-09-10
[agenda] add very basic meta tags for facebook and twitter

luto:
This addresses #122, but does not fix it completely.


rixx:
:bird: :heart:

luto:
woop! woop!

PR #129, opened 2017-09-12
CSS validator nitpicking

luto:
1. the CSS parser always returns keys in lower-case. Calling `.lower()` on them doesn't do much. Also add this to the testcase.
2. the CSS parser does not support SVG attributes like `fill`. They never survive `stylesheet.valid`. Checking for them will thus always fail, as they are never present.


rixx:
Thanks! :turtle:

PR #132, opened 2017-09-13
[submission] add missing migration for ecb8280

luto:
Just a auto-migration for ecb8280 to keep django quiet.


rixx:
I kind of dislike having lots of migrations due to the initial time they take (or having to deal with squashed migrations, which brings its own risks, so I avoid adding migrations that are functionally NOOPs such as altered help texts or choices.

PR #133, opened 2017-09-13
Delete submissions

rixx:


src/pretalx/orga/templates/orga/submission/delete.html
```diff
@@ -0,0 +1,15 @@
+{% extends "orga/submission/base.html" %}
+{% load i18n %}
+{% load url_replace %}
+
+{% block content %}
+    {% trans "Do you really want to delete this submission?" %}
+    <form method="post" action="">
+        <a class="btn btn-secondary" href="{{ submission.orga_urls.base }}">
```
luto: When a submission is deleted via the submission-list page, this button leads to the submission edit page, instead of back to the list.

luto:
The talk detail view currently has no delete-button. Is that a bug or a feature?

rixx:
Both fixed in 97ce61c5!

PR #134, opened 2017-09-13
[common][orga] link to pretalx.org instead of the github page

luto:
pretalx has got a fancy website. link to it! :)


rixx:
So it does!

Issue #42, opened 2017-05-30
Allow speaker-specific questions


rixx:
In the database, this is a pretty simple problem: Add an FK to the speaker, and add a question type that can be 'submission' or 'speaker'. The difficulty is the proper display in the orga area (probably with tags), and the display in the speaker's area.

We probably should collect questions/answers in *one* area, and mark submission specific questions as belonging to that submission.

PR #138, opened 2017-09-16
[WIP][submission] Extend model to speaker questions

rixx:
Ref #42 

TODO:

- [x] Display a second questions list in /orga
- [x] Add questions of either type
- [x] Make sure questions are asked during user submission (next to biography)
- [ ] Make sure questions are asked during submission as orga


src/pretalx/submission/models/question.py
```diff
@@ -36,6 +46,12 @@ class Question(LogMixin, models.Model):
         choices=QuestionVariant.get_choices(),
         default=QuestionVariant.STRING,
     )
+    target = models.CharField(
+        max_length=QuestionTarget.get_max_length(),
+        choices=QuestionTarget.get_choices(),
+        default=QuestionTarget.SUBMISSION,
+        verbose_name=_('question type'),
```
luto: is it a "type" or is it a "target"?

rixx: It's a "target" internally, among other things because "type" is a function in python. It is a "type" in the user interface since that is much more understandable.

luto: welp, okay :P

PR #139, opened 2017-09-17
[WIP] Review

rixx:
- [x] Honor permissions
- [x] Provide typeahead when adding reviewers
- [x] Move reviews to a better place
- [x] Add a tiled review dashboard
- [x] Add a "next unreviewed" link to smoothen reviewing workflow
- [x] Make text-only reviews work
- [x] list reviews in submission detail view
- [x] Test, test, test, and potentially forbid negative scores
- [x] Add tests
  - [x] Review workflow (submit, edit, delete, see todos)
  - [x] Average score calculation
  - [x] Permissions
  - [x] Invitations
- [x] Document the feature


src/pretalx/orga/templates/orga/review/base.html
```diff
@@ -0,0 +1,45 @@
+{% extends "orga/base.html" %}
+{% load i18n %}
+{% block headline %}
+    {% if submission %}
+        {% blocktrans with title=submission.title %}
+            Reviews for submission »{{ title }}«
+        {% endblocktrans %}
+        ({{ submission.display_speaker_names }})
+        <a href="{% url "orga:submissions.content.view" event=request.event.slug pk=submission.pk %}">
```
luto: Using `pk=` here causes a `NoReverseMatch` error; `code=` makes it work, but that's kind of backwards, as the pk is explicitly not the code. `SubmissionViewMixin` also seems like the code should be given, not the pk. Welp.

src/pretalx/orga/forms/review.py
```diff
@@ -0,0 +1,87 @@
...
+        help_text=_('The minimum score reviewers can assign'),
+    )
+    review_max_score = forms.IntegerField(
+        label=_('Maximum score'),
+        help_text=_('The maximum score reviewers can assign'),
+    )
+
+    def __init__(self, obj, *args, **kwargs):
+        super().__init__(*args, obj=obj, **kwargs)
+        minimum = int(obj.settings.review_min_score)
```
luto: Opening the review settings gives me a

> TypeError at /orga/event/pw17/settings/reviews
> int() argument must be a string, a bytes-like object or a number, not 'NoneType'

maybe because there are no defaults defined for those settings?

src/pretalx/orga/forms/review.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, event, *args, **kwargs):
+        self.event = event
+        super().__init__(*args, **kwargs)
+        min_value = int(event.settings.review_min_score or 0)
+        max_value = int(event.settings.review_max_score or 0)
+        choices = []
+        for counter in range(abs(max_value - min_value) + 1):
+            value = min_value + counter
+            name = event.settings.get(f'score_name_{value}')
```
luto: should be `review_score_name_`

luto:
Looks like a very nice addition, which was also requested by my c3w people :)

src/pretalx/orga/templates/orga/review/base.html
rixx: Yeah, parts of this PR are from the olden times™ when URLs were different. I've added a commit that uses urlman-urls instead.

rixx:
I'll add new commits incrementally for now for easier review and will restructure them before the merge.

Open questions (I'll add a todo list on top aswell):

- Do you think the UI part works? Somehow the workflow is not great yet. Part of that is the inability to actually read the submission when reviewing it, but since a submission has so many text parts, that's kind of tricky to do well.
- I think the actual reviews should be in the submission part of the UI in a separate tab just like answered questions (for example).
- There should be an easy way to grant orga members reviewing permissions aswell, not sure how to do this UI wise. Maybe suggest orga members proactively?
- Review texts should probably be shown in the list view (probably with a show/hide element) so that reading all reviews of a submission is painless.
- I'm tempted to make review texts mandatory.

src/pretalx/orga/forms/review.py
rixx: Yup, it is fixed!

src/pretalx/orga/forms/review.py
rixx: Yup, thanks.

luto:
> Do you think the UI part works? Somehow the workflow is not great yet. Part of that is the inability to actually read the submission when reviewing it, but since a submission has so many text parts, that's kind of tricky to do well.

Once the a few things like a motivational "10 more submissions to review" as well as the mentioned "save and next"-button have been added, I think it works quite well.

> I think the actual reviews should be in the submission part of the UI in a separate tab just like answered questions (for example).

Yes.

> There should be an easy way to grant orga members reviewing permissions aswell, not sure how to do this UI wise. Maybe suggest orga members proactively?

Already done. Suggesting would be nice I guess, but no necessary.

> Review texts should probably be shown in the list view (probably with a show/hide element) so that reading all reviews of a submission is painless.

yes.

> I'm tempted to make review texts mandatory.

"review texts" as in the big free-form textfield a reviewer sees or as in the description for review points? the former: no, I think that score-only reviews should work as well; the latter: forcing admins to laben their scoring system would be smart, I think.

> Do we allow changing the review score range once the first review has been submitted? This is really tricky, and I tend to just forbid it.

Changing the scale after people submitted reviews should not be possible. Maybe add a way for the admin to delete (all?) existing reviews, in case they were just playing around?

luto:
> - [ ] Make text-only reviews work

This should be fixed by b41013aa39d69962044bfa6d1906c79999ec15af.

rixx:
>> - [ ] Make text-only reviews work

>This should be fixed by b41013aa39d69962044bfa6d1906c79999ec15af.

Nope, not quite (and also, what was that *multiple* choicefield doing there in the first place?). 43d7b03b7a09 ought to do it though.

src/pretalx/orga/templates/orga/settings/review.html
```diff
@@ -0,0 +1,63 @@
...
+                {% if member != request.user or team|length > 1 %}
+                    <a href="{% url "orga:settings.review.delete" event=request.event.slug pk=member.pk %}" class="btn btn-sm btn-secondary">
+                        <span class="fa fa-times"></span>
+                    </a>
+                {% endif %}
+            </li>
+        {% endfor %}
+        <li><form method="post" action="{% url "orga:settings.review.add" event=request.event.slug %}">
+            <input type="text" name="nick" class="form-control typeahead" id="input-nick"/>
+            <button type="submit" class="btn btn-sm btn-primary"><span class="fa fa-check"></span></button>
```
luto: other speaker/team lists in `settings/team` and `submissions/42/speakers` use `fa-plus` as well as a class for common look and feel.

src/pretalx/orga/views/settings.py
```diff
@@ -181,7 +181,119 @@ def dispatch(self, request, event, pk):
...
+            permissions__event=self.request.event,
+        )
+        ctx['pending'] = EventPermission.objects.filter(event=self.request.event, user__isnull=True, is_reviewer=True)
+        return ctx
+
+
+class EventReviewInvite(View):
+
+    def post(self, request, event):
+        nick = request.POST.get('nick')
```
luto: This method is kind of long and has an early return in the middle, which makes it more confusing. Maybe it can be split into two parts: `_invite_existing_user()` and `_invite_new_user()`?

src/pretalx/orga/templates/orga/settings/review.html
rixx: Done.

src/pretalx/orga/views/settings.py
rixx: Done.

PR #144, opened 2017-09-19
[agenda] fix agenda table layout for long room names

luto:
The agenda table headers do not line up properly, if one of the room names is too long to fit:

image

This MR just makes the name wrap:

image

Is this a smart enough solution or do we want to expand the table? Or cut the name off? Ellipsis do not seem to work with `overflow-wrap`.


rixx:
I'd prefer wrapping over cutting, tbh. Let's go with this solution for now – it's definitely better than the status quo!

PR #145, opened 2017-09-21
[WIP] Add proper settings handling

rixx:
- [x] Use proper settings structure
- [ ] Add tests?
- [x] Add example settings file
- [x] Add documentation
- [x] Add compatibility for older structure

### DIFF
--- a/src/pretalx/settings.py
+++ b/src/pretalx/settings.py
@@ -1,8 +1,7 @@
 import configparser
 import os
-from contextlib import suppress
-
 import sys
+from contextlib import suppress
 from urllib.parse import urlparse
 
 from django.contrib.messages import constants as messages  # NOQA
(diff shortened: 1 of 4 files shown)


src/pretalx/settings.py
```diff
@@ -11,28 +10,98 @@
...
-encoding='utf-8')
+    config.read([
+        '/etc/pretalx/pretalx.cfg',
+        os.path.expanduser('~/.pretalx.cfg'),
+        'pretalx.cfg',
+    ], encoding='utf-8')
+
+env_config = {
+    'filesystem': {
+        'data': os.getenv('PRETALX_DATA_DIR'),
```
luto: Most of these env variables are introduced in this MR, right? Is there a reason to not make them match the config file options? e.g. `[filesystem] data` => `PRETALX_FILESYSTEM_DATA`.

src/pretalx/settings.py
```diff
@@ -11,28 +10,98 @@
 
 
 config = configparser.RawConfigParser()
+config.read_dict({
+    'filesystem': {
+        'base': os.path.dirname(os.path.dirname(__file__)),
+    },  # defaults depend on the data dir and need to be set once the data dir is fixed
+    'site': {
+        'debug': 'runserver' in sys.argv,
+        'url': 'http://localhost',
```
luto: does this somehow take the port into account later on?

rixx: Nope, but I'd like to tackle that problem later if at all

src/pretalx/settings.py
rixx: Nope, most of these options have been around already, which is the reason for the somewhat inconsistent naming.

PR #182, opened 2017-10-17
fixed 404 link in dashboard (edit cfp)

tabascoeye:
the link "Write a cfp" in the dashboard lead to a 404.
used new url definition since it looks more similar to the other
orga_urls. Don't know if that is how it is supposed to be or another value would have been better.

good question ;o)
- [X] All new and existing tests passed.

### DIFF
--- a/src/pretalx/event/models/event.py
+++ b/src/pretalx/event/models/event.py
@@ -138,6 +138,8 @@ class Event(LogMixin, models.Model):
         create = '/orga/event/new'
         base = '/orga/event/{self.slug}'
         cfp = '{base}/cfp'
+        view_cfp = '{cfp}/text'
+        edit_cfp = '{cfp}/text/edit'
         users = '{base}/users'
         mail = '{base}/mails'
         send_mails = '{mail}/send'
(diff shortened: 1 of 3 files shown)


src/pretalx/event/models/event.py
```diff
@@ -138,6 +138,8 @@ class orga_urls(Urls):
         create = '/orga/event/new'
         base = '/orga/event/{self.slug}'
         cfp = '{base}/cfp'
+        view_cfp = '{cfp}/text'
```
rixx: These are accessible at `event.cfp.urls.text` and `event.cfp.urls.edit_text` at the moment.

src/pretalx/orga/templates/orga/event/dashboard.html
```diff
@@ -16,7 +16,7 @@
                     <li><a href="{{ request.event.orga_urls.team_settings }}">
                             {% trans "Gather your team" %}
                     </a></li>
-                    <li><a href="{{ request.event.orga_urls.cfp }}">
+                    <li><a href="{{ request.event.orga_urls.edit_cfp }}">
```
rixx: So this URL should be `request.event.cfp.urls.text` instead.

tabascoeye: yeah I figured it would work that way. Just thought it odd to have all "orga_urls." except for the cfp. But yeah..

rixx:
Closed in bdc8668.

Issue #176, opened 2017-10-13
add tooltip explaning do-not-record icon


luto:
In the public schedule view do-not-record is currently shown as a 🚫 symbol. This should but clear to many people, but maybe not to all. We could add a tooltip to make it clearer in a non-intrusive way.

rixx:
We already have the hover CSS tooltip and the regular html tooltip. We should probably add this information to either of these, not add a third one. :wink:

tabascoeye:
@luto: agree to close this after #184 was merged?

rixx:
Definitely a closed issue :)

PR #184, opened 2017-10-17
more visible icon for the "do not record" flag

tabascoeye:
This PR closes issue #176. It does so by adding a "hover" tooltip to the icon and changes it by stacking the font-awesome "video camera" icon with the "banned" icon on top.


rixx:
Thanks, looks much better than the previous version!

PR #187, opened 2017-10-22
[WIP] Finish speaker avatar feature

rixx:
- [x] add and test speaker page
- [x] add optional gravatar avatar


src/pretalx/static/cfp/js/profile.js
```diff
@@ -0,0 +1,10 @@
+$('#id_get_gravatar').click(() => {
+    if ($('#id_get_gravatar').is(':checked')) {
+        $('.user-avatar-display').slideUp();
```
luto: When a speaker checks this box the file upload vanishes, but nothing else happens. It would be nice to load the gravatar picture, so the speaker knows, if the box actually did something.

rixx: Done.

PR #198, opened 2017-10-28
163 inactive questions

rixx:


src/pretalx/cfp/forms/submissions.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, *args, **kwargs):
 
         super().__init__(*args, **kwargs)
 
-        queryset = self.event.questions.all()
+        queryset = self.event.questions.filter(active=True)
```
luto: Is it maybe possible to [use `QuestionManager.use_for_related_fields`](https://stackoverflow.com/a/7490852/2486196), so related managers also use `QuestionManager` (active questions only) instead of the default one returning all questions?

rixx: Ah, no, the change in this line is actually unnecessary, as I (later) made the `active_question` manager the default manager, as visible [here](https://github.com/pretalx/pretalx/blob/8c1f4716f392181d6c9df16b55ba19a56a360eea/src/pretalx/submission/models/question.py#L92)

luto: https://github.com/pretalx/pretalx/commit/2a426aa488f8d311b91307cb79800c893203edc3 🎉

src/pretalx/orga/views/cfp.py
```diff
@@ -164,7 +164,9 @@ def dispatch(self, request, *args, **kwargs):
             question.log_action('pretalx.question.delete', person=self.request.user, orga=True)
             messages.success(request, _('The question has been deleted.'))
         except ProtectedError:
-            messages.error(request, _('You cannot delete a question that has already been answered.'))
+            question.active = False
+            question.save()
+            messages.error(request, _('You cannot delete a question that has already been answered. We have deactivated the question instead.'))
```
luto: This works, but leaves the UI unchanged. The question still displays a Trash icon. Pressing it causes pretalx to crash:

```
File "pretalx/src/pretalx/orga/views/cfp.py" in dispatch
  173.         question = self.request.event.questions.get(pk=self.kwargs.get('pk'))

File "pretalx/venv/lib/python3.6/site-packages/django/db/models/manager.py" in manager_method
  85.                 return getattr(self.get_queryset(), name)(*args, **kwargs)

File "pretalx/venv/lib/python3.6/site-packages/django/db/models/query.py" in get
  380.                 self.model._meta.object_name

Exception Type: DoesNotExist at /orga/event/pw17/cfp/questions/2/delete
Exception Value: Question matching query does not exist.
```

src/pretalx/orga/templates/orga/cfp/question_form.html
```diff
@@ -15,6 +15,9 @@
         {% bootstrap_field form.variant layout='horizontal' %}
         {% bootstrap_field form.help_text layout='horizontal' %}
         {% bootstrap_field form.required layout='horizontal' %}
+        {% if action != 'create' %}
+            {% bootstrap_field form.active layout='horizontal' %}
```
luto: given the impact, a "deactivate"/"activate"-button would be nicer here.

rixx: Moved to #246

PR #199, opened 2017-10-28
56 email on submission

rixx:


src/pretalx/cfp/views/wizard.py
```diff
@@ -143,6 +136,23 @@ def done(self, form_list, form_dict, **kwargs):
...
+            if self.request.event.settings.mail_on_new_submission:
+                MailTemplate(
+                    event=self.request.event,
+                    subject=_('[{event}] New submission!').format(event=self.request.event.slug),
+                    text=self.request.event.settings.mail_text_new_submission,
+                ).to_mail(
+                    user=user, event=self.request.event, context=template_context_from_submission(sub),
+                    skip_queue=True, locale=self.request.event.locale
+                )
+        except SendMailException:
```
luto: Split this `try`, in case there is a problem specific to one template or mail address, we allow the other one to go through. Probably unlikely, but worth a shot, I think.

src/pretalx/mail/context.py
```diff
@@ -46,9 +46,8 @@ def template_context_from_submission(submission):
         ),
         'event_name': submission.event.name,
         'submission_title': submission.title,
-        'submission_url': build_absolute_uri(
-            'cfp:event.user.submission.edit',
-            kwargs={'event': submission.event.slug, 'code': submission.code}
-        ),
+        'submission_url': submission.urls.user_base.full(scheme='https'),
```
luto: optional side-note: `PRETALX_SITE_URL` and similar settings allow a full base URL with protocol to be specified. While I agree that everything should be https://, because 2017, this does not honor the settings.

src/tests/functional/cfp/test_submission_wizard.py
```diff
@@ -108,6 +111,7 @@ def test_wizard_new_user(self, event, question, client):
         assert user.email == 'testuser@example.org'
         assert user.name == 'Jane Doe'
         assert user.profiles.get(event=event).biography == 'l337 hax0r'
+        assert len(djmail.outbox) == 2  # user email plus orga email
```
luto: maybe check the subject so we know the right templates are being used for each mail?

src/pretalx/cfp/views/wizard.py
```diff
@@ -143,6 +136,23 @@ def done(self, form_list, form_dict, **kwargs):
                     answer.answer = value
                 answer.save()
 
+        try:
+            sub.event.ack_template.to_mail(
+                user=user, event=self.request.event, context=template_context_from_submission(sub),
+                skip_queue=True, locale=user.locale
+            )
+            if self.request.event.settings.mail_on_new_submission:
+                MailTemplate(
```
luto: this is sent to the speaker, instead of the orga.

src/pretalx/mail/context.py
rixx: I don't quite agree – we have strong warnings against http sites in our documentation, and development urls are invalid in any case atm because we don't try to add the correct port number.

I'll disregard this for now because I'll add a custom Url class in a few days which will have all the `full()` logic in one place and can be clever enough to figure this out correctly.

Issue #67, opened 2017-08-23
Add ical export


rixx:
- [x] Add ical export for single event
- [x] Add ical export for all events
- [x] Add ical export for all events of a speaker: https://github.com/pretalx/pretalx/pull/304
- [x] Add ical url for single event
- [x] Add ical url for all events
- [ ] Add ical export for group of events
- [ ] Add ical url for group of events
- [ ] Add UI mode to define a group of events

rixx:
This, of course, needs a lot of UI considerations – I'm not really sure how you'd switch to that mode and back again, except via a button in the export dropdown.

PR #202, opened 2017-10-29
ical export

rixx:
Improve #67 – add an ical export for single events and extract necessary methods for further steps.

### DIFF
--- a/src/pretalx/agenda/templates/agenda/talk.html
+++ b/src/pretalx/agenda/templates/agenda/talk.html
@@ -27,6 +27,9 @@
             </a>
         {% endif %}
         <small>
+            <a href="{{ talk.urls.ical }}">
+                <span class="fa fa-calendar"></span>
+            </a>
             {{ talk.slot.start.date|date:"c" }}, {{ talk.slot.start|date:"H:i" }}–{{ talk.slot.end|date:"H:i" }}, {{ talk.slot.room.name }}
         </small>
     </h3>
--- a/src/pretalx/agenda/views/schedule.py
+++ b/src/pretalx/agenda/views/schedule.py
@@ -239,14 +239,11 @@ class SingleICalView(EventPageMixin, DetailView):
     slug_field = 'code'
 
     def get(self, request, event, **kwargs):
-        schedule = self.get_object()
+        talk = self.get_object().slots.get(schedule=self.request.event.current_schedule)
         netloc = urlparse(settings.SITE_URL).netloc
 
         cal = vobject.iCalendar()
-        cal.add('prodid').value = '-//pretalx//{}//'.format(netloc)
-        creation_time = datetime.now(pytz.utc)
-
-        talk = self.get_object()
+        cal.add('prodid').value = '-//pretalx//{}//{}'.format(netloc, talk.submission.code)
         talk.build_ical(cal)
 
         resp = HttpResponse(cal.serialize(), content_type='text/calendar')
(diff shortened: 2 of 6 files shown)


src/pretalx/agenda/views/schedule.py
```diff
@@ -141,21 +140,7 @@ def get(self, request, event, **kwargs):
...
-            vevent.add('uid').value = 'pretalx-{}-{}@{}'.format(
-                request.event.slug, talk.submission.code,
-                netloc
-            )
-
-            vevent.add('dtstart').value = talk.start.astimezone(tz)
-            vevent.add('dtend').value = talk.end.astimezone(tz)
-            vevent.add('description').value = talk.submission.abstract or ""
-            vevent.add('url').value = talk.submission.urls.public.full(scheme='https')
+            talk.build_ical(cal, creation_time=creation_time, netloc=netloc)
```
luto: `creation_time` already defaults to `datetime.now(pytz.utc)`. Actually, it is seemingly never set to something other than that. In case there isn't some hidden meaning I don't get: think about removing the parameter or at least don't pass it over here.

src/pretalx/schedule/models/slot.py
```diff
@@ -71,3 +74,22 @@ def copy_to_schedule(self, new_schedule, save=True):
...
+        vevent.add('location').value = str(self.room.name)
+        vevent.add('uid').value = 'pretalx-{}-{}@{}'.format(
+            self.submission.event.slug, self.submission.code,
+            netloc
+        )
+
+        vevent.add('dtstart').value = self.start.astimezone(tz)
+        vevent.add('dtend').value = self.end.astimezone(tz)
+        vevent.add('description').value = self.submission.abstract or ""
+        vevent.add('url').value = self.submission.urls.public.full(scheme='https')
```
luto: Use the scheme out of `settings.SITE_URL` here. Everything should be https:// in 2017, but this is still a settings which should be honored here. Dev instances are http most of the time, for example.

src/pretalx/agenda/templates/agenda/talk.html
```diff
@@ -27,6 +27,9 @@ <h3 class="talk-title">
             </a>
         {% endif %}
         <small>
+            <a href="{{ talk.urls.ical }}">
+                <span class="fa fa-calendar"></span>
```
luto: The icon just looks like a "here comes the date!"-indicator for the following text and has roughly the same color as the text. Even after skimming this, I needed to go back to the code to find out how to actually get the ical file in the UI. Sadly ical doesn't seem to have any kind of icon, so I don't really know what do to here or where to move the link..

luto: Maybe I am just blind and it is fine. Time will show.

src/pretalx/agenda/views/schedule.py
rixx: When generating a lot of talks for *one* ical, I want a consistent creation time (and only call `now()` once, no reason to have `n` method calls when you can have just one.

src/pretalx/schedule/models/slot.py
rixx: See other comment. Will happen in a site-wide class instead of in six (or so) separate places.

src/pretalx/agenda/templates/agenda/talk.html
rixx: We could make it a button?

rixx: I made it a button: 89452e96

PR #203, opened 2017-10-29
Add missing action logs and display them

rixx:
Displays log entries in the user's frontend.

2017-10-30-001641_1467x805_scrot

### DIFF
--- a/src/pretalx/orga/views/submission.py
+++ b/src/pretalx/orga/views/submission.py
@@ -77,6 +77,7 @@ class SubmissionAccept(SubmissionViewMixin, View):
 
         try:
             submission.accept(person=request.user)
+            submission.log_action('pretalx.submission.accept', person=request.user, orga=True)
             messages.success(request, _('The submission has been accepted.'))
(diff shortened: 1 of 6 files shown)


src/pretalx/orga/views/submission.py
```diff
@@ -90,6 +91,7 @@ def dispatch(self, request, *args, **kwargs):
 
         try:
             submission.confirm(person=request.user, orga=True)
+            submission.log_action('pretalx.submission.confirm', person=request.user, orga=True)
```
luto: There is also `pretalx.submission.confirmation`.. sooo is there any easy way to merge them now?

rixx: Ehh, no, but since this was not in a release, we'll just silently change it and hope for the best. Done in 1da7893c.

rixx:
That is true, I was planning on tackling those before the next release, which is when I'll go through translations again aswell (because lots of "I translated a string" commits are kinda annoying, and the changesets are hard to read).

PR #218, opened 2017-11-03
Optionally hide schedule

rixx:


rixx:
> Merging #218 into master will decrease coverage by 0.03%.
> The diff coverage is 100%.

I think I'll merge it then.

src/pretalx/agenda/templates/agenda/base.html
```diff
@@ -13,4 +13,19 @@
 {% endblock %}
 
 {% block content %}
+{% if request.event.settings.show_schedule or request.is_orga %}
```
luto: It works great for hiding the talk and agenda pages, but the button on the front page leading to them is still there. Clicking it shows a blank page.

rixx: I'll put the fix for this in the permissions branch, so that it does not get lost: 42d5b9db

PR #219, opened 2017-11-03
[orga] Pre-select question type according to button clicked

rixx:


luto:
yay for mini-features! :)

PR #220, opened 2017-11-03
[orga] Validate mail template placeholders onsave

rixx:


src/pretalx/orga/forms/mails.py
```diff
@@ -2,11 +2,28 @@
...
 
 class MailTemplateForm(ReadOnlyFlag, forms.ModelForm):
 
+    def __init__(self, *args, event=None, **kwargs):
+        self.event = event
+        super().__init__(*args, **kwargs)
+
+    def clean_text(self):
+        text = self.cleaned_data['text']
+        if self.instance and self.instance.id and self.instance in [t for t in self.instance.event.fixed_templates if t != self.event.update_template]:
```
luto: Maybe it's juts me, but: What does this line do? The condition is kind of long and I don't get what the last expression filters. Maybe move it into a dedicated, named variable to clear things up?

rixx: Improved upon in fff4bcbe, thanks!

PR #230, opened 2017-11-05
Add celery settings and docs

### LINKED ISSUE #38: Actually use celery and set it up

rixx:
Closes #38


src/pretalx/common/management/commands/rebuild.py
```diff
@@ -0,0 +1,11 @@
+from django.core.management import call_command
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = "Rebuild static files and language files"
```
luto: 9/10, very handy, would call command again.

PR #232, opened 2017-11-07
Introduce a rule-based permissions system

### LINKED ISSUE #78: Plan a permissions system
pretalx will, in the end, work as a role based permissions system, and now is probably the best time to decide on how/where/which roles to distribute.

After researching this for a bit, [rules](https://github.com/dfunckt/django-rules) seems the way to go.

rixx:
Closes #78 

Tbd (but not necessarily before a merge): Use new permission rules in templates aswell, e.g. to determine if a certain menu or button should be shown.


src/pretalx/common/permissions.py
```diff
@@ -0,0 +1,9 @@
+from django.http import Http404
+from rules.contrib.views import PermissionRequiredMixin
+
+
+class PermissionRequired(PermissionRequiredMixin):
```
luto: does this  form mixin belong in `permissions.py`? This file seems to be reserved for predicates and rules in all other cases.

src/pretalx/submission/permissions.py
```diff
@@ -0,0 +1,68 @@
...
+
+rules.add_perm('submission.withdraw_submission', can_be_withdrawn & is_speaker)
+rules.add_perm('submission.reject_submission', can_be_rejected & is_orga)
+rules.add_perm('submission.accept_submission', can_be_accepted & is_orga)
+rules.add_perm('submission.confirm_submission', can_be_confirmed & (is_speaker | is_orga))
+rules.add_perm('submission.cancel_submission', can_be_canceled & (is_speaker | is_orga))
+rules.add_perm('submission.unconfirm_submission', can_be_unconfirmed & (is_speaker | is_orga))
+rules.add_perm('submission.remove_submission', can_be_removed & is_orga)
+rules.add_perm('submission.edit_submission', (can_be_edited & is_speaker) | is_orga)
+rules.add_perm('submission.review_submission', is_reviewer & can_review_submission)
```
luto: just to confirm: being orga is not enough to review a talk, you must be a reviewer as well?

src/pretalx/submission/permissions.py
```diff
@@ -0,0 +1,68 @@
...
+    return SubmissionStates.ACCEPTED in SubmissionStates.valid_next_states.get(obj.state, []) and obj.state == SubmissionStates.CONFIRMED
+
+
+@rules.predicate
+def can_be_edited(user, obj):
+    return obj.editable
+
+
+@rules.predicate
+def can_review_submission(user, obj):
```
luto: the _predicate_ here would be `is_not_involved` or `is_not_speaker`, wouldn't it? Or, maybe even better: use the negated `~is_speaker` in `submission.review_submission` and drop this predicate.

src/pretalx/submission/permissions.py
```diff
@@ -0,0 +1,68 @@
...
+rules.add_perm('submission.reject_submission', can_be_rejected & is_orga)
+rules.add_perm('submission.accept_submission', can_be_accepted & is_orga)
+rules.add_perm('submission.confirm_submission', can_be_confirmed & (is_speaker | is_orga))
+rules.add_perm('submission.cancel_submission', can_be_canceled & (is_speaker | is_orga))
+rules.add_perm('submission.unconfirm_submission', can_be_unconfirmed & (is_speaker | is_orga))
+rules.add_perm('submission.remove_submission', can_be_removed & is_orga)
+rules.add_perm('submission.edit_submission', (can_be_edited & is_speaker) | is_orga)
+rules.add_perm('submission.review_submission', is_reviewer & can_review_submission)
+rules.add_perm('submission.view_submission', is_speaker | is_orga | is_reviewer)
+rules.add_perm('submission.add_speaker', is_speaker | is_orga)
```
luto: This is used for both adding and removing, not just adding. Maybe call it `submission.edit_speaker_list` or `.edit_speakers`?

src/pretalx/orga/views/settings.py
```diff
@@ -33,7 +41,11 @@ def dispatch(self, request, *args, **kwargs):
         if self._action == 'create':
             if not request.user.is_anonymous and not request.user.is_superuser:
                 raise PermissionDenied()
-        return super().dispatch(request, *args, **kwargs)
+        ret = super().dispatch(request, *args, **kwargs)
+        if self._action != 'create':
```
luto: There is no open transaction here. If I am not mistaken, any changes caused by the `dispatch()` above will not be reverted when we raise an Exception below. So the users sees an error page, but the changes are saved nonetheless. A quick and dirty test seems to confirm that.

This check does not seem to depend on the result of `dispatch()`, so move the if above the call, just to be sure?

src/pretalx/orga/permissions.py
```diff
@@ -0,0 +1,31 @@
...
+rules.add_perm('orga.purge_mails', is_orga)
+rules.add_perm('orga.view_review_dashboard', is_orga | is_reviewer)
+rules.add_perm('orga.view_reviews', is_reviewer)
+rules.add_perm('orga.remove_review', is_reviewer)
+rules.add_perm('orga.view_schedule', is_orga)
+rules.add_perm('orga.release_schedule', is_orga)
+rules.add_perm('orga.edit_schedule', is_orga)
+rules.add_perm('orga.schedule_talk', is_orga)
+rules.add_perm('orga.view_room', is_orga)
+rules.add_perm('orga.edit_room', is_orga)
```
luto: This is never actually used. Instead, `RoomDetail`, which is currently part of the general event settings tabs, relies on `orga.change_settings`. But since `orga.view_room` is used, `.edit_room` should be, too.

src/pretalx/orga/permissions.py
```diff
@@ -0,0 +1,31 @@
...
+rules.add_perm('orga.remove_question', is_orga)
+rules.add_perm('orga.view_submission_type', is_orga)
+rules.add_perm('orga.edit_submission_type', is_orga)
+rules.add_perm('orga.remove_submission_type', is_orga)
+rules.add_perm('orga.view_mails', is_orga)
+rules.add_perm('orga.send_mails', is_orga)
+rules.add_perm('orga.purge_mails', is_orga)
+rules.add_perm('orga.view_review_dashboard', is_orga | is_reviewer)
+rules.add_perm('orga.view_reviews', is_reviewer)
+rules.add_perm('orga.remove_review', is_reviewer)
```
luto: This means that a reviewer can remove reviews written by a different reviewer, right? I think reviewers should only be able to delete their own reviews. Reviews by other people should only be deletable by orga, or not at all.

src/pretalx/orga/views/mails.py
```diff
@@ -155,8 +183,12 @@ def form_valid(self, form):
         return super().form_valid(form)
 
 
-class TemplateList(TemplateView):
+class TemplateList(PermissionRequired, TemplateView):
     template_name = 'orga/mails/template_list.html'
+    permission_required = 'orga.view_mails'
```
luto: Mail templates are important enough to deserve their own set of permissions, I think. So this should be `orga.view_mail_templates`, `TemplateDetail` should be `orga.edit_mail_template` (+ hide the edit button), and `TemplateDelete` should also get `orga.edit_mail_template` (+ hide the delete button).

src/pretalx/orga/views/mails.py
```diff
@@ -43,7 +52,11 @@ def get_queryset(self):
         return qs
 
 
-class OutboxSend(View):
+class OutboxSend(PermissionRequired, View):
+    permission_required = 'orga.view_mails'
```
luto: This seems to actually send mails. So `orga.send_mails` like the `SendMail` view, which just queues mail? I am not sure whether queuing or actually sending mails or both should need `orga.send_mails`.

src/pretalx/orga/views/mails.py
```diff
@@ -69,7 +82,13 @@ def dispatch(self, request, *args, **kwargs):
         return redirect(self.request.event.orga_urls.outbox)
 
 
-class OutboxPurge(View):
+class OutboxPurge(PermissionRequired, View):
+    permission_required = 'orga.view_mails'
```
luto: There is `orga.purge_mails`, which seems more fitting here?

src/pretalx/common/permissions.py
rixx: It's a view mixin; I'll move it to `common.views`.

rixx: Things in `pretalx.common` usually contain more generalized forms of things – in `common.views`, there are mixins for views (or very generic base classes), but in `common.models`, we have one actual model. I'd be open to a `pretalx.common` refactoring that moves *all* mixins to `pretalx.common.mixins.views`, `pretalx.common.mixins.forms`, etc., but I don't think it's worth changing in just this PR.

src/pretalx/submission/permissions.py
rixx: Exactly (which is also the way this currently works). At the moment an orga member can just give themselves reviewer permission, but at least this will be transparently logged.

The thought behind this is that reviews are rather personal, since they can contained detailed criticism of the speaker's content aswell as her background. Leaving this open to only reviewers permits us to further restrict access in the future (e.g. by adding a role of `review-orga` who are the only ones able to invite reviewers).

src/pretalx/submission/permissions.py
rixx: Excellent point, this is much more readable.

src/pretalx/submission/permissions.py
rixx: Thanks, definitely worth renaming!

src/pretalx/orga/views/settings.py
rixx: Yup, that's right.

src/pretalx/orga/permissions.py
rixx: Good point, fixed.

src/pretalx/orga/permissions.py
rixx: Fixed.

src/pretalx/orga/views/mails.py
rixx: Queueing should be `'edit_mails'`, while sending requires `'send_mails'`.

src/pretalx/orga/views/mails.py
rixx: Done.

src/pretalx/orga/views/mails.py
rixx: Yup.

PR #263, opened 2017-11-13
Fix mail problem, closes #247

### LINKED ISSUE #247: Mail config doesn't work
The fallback/default config is not used, even when an event's mail config is untouched.

rixx:


luto:
This is a breaking change not a bugfix, in case the config relies on the old default value and the server does not support unencrypted connections. It can also cause security issues for servers that do support both unencrypted and encrypted traffic.

Please document it as such in the changelog.

rixx:
Done, thank you.

PR #264, opened 2017-11-14
Add email logging option

rixx:
Okay, I maybe refactored our settings in the progress. So what!

### DIFF
--- a/src/pretalx/settings.py
+++ b/src/pretalx/settings.py
@@ -25,6 +25,10 @@ config.read_dict({
     'filesystem': {
         'base': os.path.dirname(os.path.dirname(__file__)),
     },  # defaults depend on the data dir and need to be set once the data dir is fixed
+        # 'base'
+        # 'logs': 'logs',
+        # 'media': 'media'
+        # 'static': 'static.dist'
     'site': {
         'debug': 'runserver' in sys.argv,
         'url': 'http://localhost',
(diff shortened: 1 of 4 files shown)


src/pretalx/settings.py
```diff
@@ -8,12 +8,23 @@
...
 
 config, config_files = build_config()
 
-# File system and directory settings
+##
+# This settings file is rather lengthy. It follows this structure:
+# Directories, Apps, Url, Security, Databases, Logging, Email, Caching (and Sessions)
+# I18n, Auth, Middleware, Templates and Staticfiles, External Apps
+#
+# Search for "## {AREA} SETTINGS" to navigate this file
```
luto: 🎉

src/pretalx/settings.py
```diff
@@ -405,6 +421,9 @@ def reduce_dict(data):
 
 MESSAGE_STORAGE = 'django.contrib.messages.storage.session.SessionStorage'
 
+emails = config.get('logging', 'email', fallback='').split(',')
+ADMIN = [(email, email) for email in emails if email]
```
luto: since we split those values into a list, trimming them would be nice in case someone sets a value like `email@one.com, admin@google.com`.

src/pretalx/settings.py
```diff
@@ -452,6 +471,12 @@ def reduce_dict(data):
         }
     },
 }
+email_level = config.get('logging', 'email_level', fallback='ERROR') or 'ERROR'
```
luto: in which case do you expect the `or 'ERROR'` to be executed? someone sets a value (and thus circumvents `fallback=`), buuut the value is falsy?

Issue #277, opened 2017-11-19
export HTML schedule on schedule release & intervals


luto:
Static schedule exports can currently only be triggered manually via a management command. But we know when the data changes so looking at #275 we should do that automatically, if the user wishes us to do so.

- [x] generate export on schedule release #297
- [ ] maybe generate in regular intervals
- [ ] maybe generate on talk changes, e.g. the title changes: is there a way to catch bluk updates?

rixx:
There is no good way to catch bulk updates, but we don't really do those. Instead on updating on talk changes directly, how about storing a dirty flag that will be reset once the schedule has been regenerated?

rixx:
The HTML export also needs to be regenerated when the organiser uploads a changed CSS or logo.

rixx:
We now have a good way to do this!

```python
event.cache.set('rebuild_schedule_export', True, None)
```

will have to be added when

- [ ] A talk is confirmed
- [ ] A talk is unconfirmed
- [ ] A talk changes in a visible way
- [ ] A speaker changes in a visible way
- [ ] Related data to a talk or a speaker (question answers!)

rixx:
I decided that it doesn't hurt to be careful, so this flag is set on any submission save and speaker save, currently (as long as the schedule is configured to be exported on release). I think it would be way harder to make sure a change does/does not affect the schedule than to just trigger a regeneration of the schedule export periodically. I've limited the automatic regeneration to 30 minutes to be on the safe side in terms of DoS.

PR #297, opened 2017-12-15
Orga schedule export

### LINKED ISSUE #275: offer static schedule export in orga area & API
As of #269 there is a management command to export for all public schedule pages as static HTML to `data/htmlexport` (or a different path, as configured). To make this accessible to non server admins, there should be a button to generate an export on-demand in the orga area.

- [x] add a celery task to generate schedule in background
- [x] add button in orga area to trigger celery task
- [x] add optional zipping to exporter
- [x] offer generated zips for download in orga area
- [x] integrate into the readonly API

luto:
Fixes #275 
Refs #277

## How Has This Been Tested?

By frantically clicking around. Real tests will follow.

## Checklist

- [ ] ~~I have updated the documentation accordingly.~~

## TODO

- [x] if a schedule is exported before the first version is published, it is publicly accessible. That's probably not smart.
- [x] make ScheduleExportTriggerView only accept POSTs
- [x] add setting to not always export schedule on release
- [x] tests: do not export normally
- [x] tests: add test for html export on schedule release
- [x] make export view in orga pretty
- [x] give the management command the ability to zip export
- [x] test `export_schedule_html` celery task
- [x] make celery task use kwargs, instead of args
- [x] test zipping
- [x] test `ScheduleExportTriggerView`
- [x] fix schedule export tests, as they probably fail now
- [x] test zip download

### DIFF
--- a/src/pretalx/event/models/event.py
+++ b/src/pretalx/event/models/event.py
@@ -139,6 +139,7 @@ class Event(LogMixin, models.Model):
         ical = '{schedule}.ics'
         feed = '{schedule}/feed.xml'
         location = '{schedule}/location'
+        html_export = f'{settings.MEDIA_URL}/{{self.slug}}/schedule_{{self.slug}}.zip'
 
     class orga_urls(EventUrls):
         create = '/orga/event/new'
--- a/src/pretalx/event/models/event.py
+++ b/src/pretalx/event/models/event.py
@@ -169,6 +169,8 @@ class Event(LogMixin, models.Model):
         invite_reviewer = '{review_settings}/add'
         new_room = '{room_settings}/new'
         schedule = '{base}/schedule'
+        schedule_export = '{base}/schedule/export'
+        schedule_export_trigger = '{base}/schedule/export/trigger'
         release_schedule = '{schedule}/release'
         reset_schedule = '{schedule}/reset'
         toggle_schedule = '{schedule}/toggle'
--- a/src/pretalx/event/models/event.py
+++ b/src/pretalx/event/models/event.py
@@ -13,7 +13,7 @@ from i18nfield.fields import I18nCharField, I18nTextField
 
 from pretalx.common.mixins import LogMixin
 from pretalx.common.models.settings import settings_hierarkey
-from pretalx.common.urls import EventUrls
+from pretalx.common.urls import EventUrls, get_base_url
 
 SLUG_CHARS = 'a-zA-Z0-9.-'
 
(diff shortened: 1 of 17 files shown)


src/pretalx/event/models/event.py
```diff
@@ -194,6 +197,10 @@ def named_locales(self) -> list:
         enabled = set(self.locale_array.split(","))
         return [a for a in settings.LANGUAGES_NATURAL_NAMES if a[0] in enabled]
 
+    @property
+    def url_html_export(self) -> str:
```
rixx: `html_export_url` would sound more natural, I think.

src/pretalx/event/models/event.py
```diff
@@ -139,6 +139,7 @@ class urls(EventUrls):
         ical = '{schedule}.ics'
         feed = '{schedule}/feed.xml'
         location = '{schedule}/location'
+        html_export = f'{settings.MEDIA_URL}/{{self.slug}}/schedule_{{self.slug}}.zip'
```
rixx: Does this work? I think urls should contain single `{}` always, never `{{ }}`

src/pretalx/schedule/models/schedule.py
```diff
@@ -72,6 +73,8 @@ def freeze(self, name, user=None, notify_speakers=True):
         with suppress(AttributeError):
             del wip_schedule.event.current_schedule
 
+        export_schedule_html.apply_async(args=(self.event.id,))
```
rixx: I'd prefer if we'd pass all celery task arguments as `kwargs` always.

luto: sure thing, added to TODOs.

src/pretalx/event/models/event.py
luto: Since this is a format-string `{` is handled by the formatter, `{{` becomes `{` and is handled by urlman, which may not be the prettiest thing. That line will probably have to go anyways, because storing the export in media_root leaks it at a point at which it may not be public yet.

src/pretalx/schedule/models/schedule.py
luto: ✅

src/pretalx/event/models/event.py
luto: ✅

src/pretalx/event/models/event.py
luto: ✅

luto:
All planned features are done and tests are now green like a Christmas tree :) Feel free to point out anything that's still unclear or not pretty enough. I'll respond faster this time.

rixx:
Thank you!

PR #304, opened 2018-01-01
add ical file for all talks of a speaker

luto:
Refs #67 

## How Has This Been Tested?

Using the automated tests provided, as well as clicking the new button in the speaker profile manually and importing the resulting file in my calendar.

image

## Checklist

### DIFF
--- a/src/pretalx/person/models/profile.py
+++ b/src/pretalx/person/models/profile.py
@@ -30,6 +30,7 @@ class SpeakerProfile(LogMixin, models.Model):
 
     class urls(EventUrls):
         public = '{self.event.urls.base}/speaker/{self.user.code}'
+        talks_ical = '{self.urls.public}/talks.ical'
 
     @cached_property
     def code(self):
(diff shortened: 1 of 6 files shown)


src/pretalx/person/models/profile.py
```diff
@@ -30,6 +30,7 @@ class SpeakerProfile(LogMixin, models.Model):
 
     class urls(EventUrls):
         public = '{self.event.urls.base}/speaker/{self.user.code}'
+        talks_ical = '{self.urls.public}/talks.ical'
```
rixx: I suppose it would make more sense to put the speaker code or speaker name here, in addition. Otherwise you'll end up with talks.ical, talks (2).ical, etc.

Also, we use .ics in other places, if memory serves, so I'd prefer this for consistency's sake.

luto: The downloaded filename contains both the slug and a (cleaned) speaker name: https://github.com/pretalx/pretalx/pull/304/files#diff-3c2419007a707ff9150a067e5d3c3b57R62

luto: ical => ics  fixed. :)

Issue #372, opened 2018-03-05
Timezone discrepancy in CfP view


derchrisuk:
## Expected Behavior
Setting a deadline in CfP->Content, and checking 'Display deadline publicly'.
Previewing the CfP it should show the correct deadline according to the servers Timezone.

## Current Behavior
Previewing the CfP shows a date for the deadline off by 7 hours.
it does show the correct TZ though, which is 'Europe/Berlin'

## Possible Solution
Checking Django config, it takes 'Americas/Chicago' as a default timezone
https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-TIME_ZONE

## Steps to Reproduce (for bugs)
1. Set deadline in CfP->Content
2. Check 'Display deadline publicly'

## Your Environment
* Version used: 0.4.1
* Operating System and version (desktop or mobile): Ubuntu stretch/sid, TZ Europe/Berlin
* Link to your project: not public

derchrisuk:
Tried adding this to settings.py did not help
`TIME_ZONE = 'Europe/Berlin'`

rixx:
PR #373 fixes this issue for me, can you confirm the fix, @derchrisuk?

derchrisuk:
What is the best way to test this using the pip install approach?

derchrisuk:
@rixx I can confirm after copying over the changes.

rixx:
Thank you!

PR #373, opened 2018-03-05
[common] Refs #372 -- Swap priorities of timezone selection

raphaelm:
This PR references issue #372. It does so by prioritizing event timezone over user timezone in the backend.

## How Has This Been Tested?

Manually and with a regression test. I'm not sure if it fixes the original problem of #372 as I can't reproduce that one locally.

## Checklist


rixx:
Thank you! :clock1:

PR #374, opened 2018-03-07
[WIP] Add links for objects in action logs for Submission and CfP

rahmonov:
Hey guys, does it look okay? Or did you have something else in mind?

### DIFF
--- a/src/pretalx/common/models/log.py
+++ b/src/pretalx/common/models/log.py
@@ -3,12 +3,14 @@ from django.contrib.contenttypes.models import ContentType
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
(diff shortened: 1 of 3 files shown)


rahmonov:
@rixx does it look okay? or do we need something finer-grained? like for separate links for submission speakers, reviews and etc?

rixx:
@rahmonov Apart from the failing code style tests, the only improvement I could think of would be separate methods to get an organiser area url and a public url – but that's something we could add later, aswell.

rahmonov:
@rixx fixed the style issues. I feel stupid but could not figure out "separate methods to get an organiser area url and a public url" :) Can we create a separate issue for this with a little more description?

rixx:
Sure, thank you very much!

PR #375, opened 2018-03-14
Fixed permissions for access to submissions via the organizers frontend

morisson:
Limits access to submissions details via the organizer frontend to organizers and reviewers only.

## How Has This Been Tested?
Tested on a local pretalx install, and shouldn't impact any other parts of the application.


rixx:
Thank you very much!

PR #377, opened 2018-03-15
Update docker install - Don't run as root

morisson:
* Don't run as root, run as uid 999
* Bind on port 8000
* change pycopg2 to pycopg2-binary (removes warning of future deprecation)

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,18 +14,26 @@ ENV LC_ALL C.UTF-8
 COPY docker/pretalx.bash /usr/local/bin/pretalx
 COPY src /src
 
-RUN mkdir /static && \
-    pip3 install -U pip setuptools wheel typing && \
+RUN pip3 install -U pip setuptools wheel typing && \
     pip3 install -e src/ && \
-    pip3 install django-redis pylibmc mysqlclient psycopg2 && \
+    pip3 install django-redis pylibmc mysqlclient psycopg2-binary && \
     pip3 install gunicorn && \
     chmod +x /usr/local/bin/pretalx
 
-RUN mkdir -p /data/logs /data/media
-VOLUME /data
+RUN groupadd -g 999 pretalx && \
+    useradd -r -u 999 -g pretalx -d /src pretalx  
+
+RUN mkdir -p /data/logs /data/media /src/static.dist
 
 RUN python3 -m pretalx migrate && python3 -m pretalx rebuild
 
-EXPOSE 80
+RUN chown -R 999:999 /data && \
+    chown -R 999:999 /src/static.dist
+
+USER pretalx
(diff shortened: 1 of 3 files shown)


rixx:
Thank you for this pull request! As I'm not very knowledgeable in terms of docker, I have to ask: will this update break things for existing docker users who switch to this version of the Dockerfile?

raphaelm:
> Thank you for this pull request! As I'm not very knowledgeable in terms of docker, I have to ask: will this update break things for existing docker users who switch to this version of the Dockerfile?

My guess would be: Files/directories created in the data folder that are now owned by root will be unwritable after the change. A simple recursive chown on the data folder should fix it. Otherwise I don't see breaking changes.

morisson:
Yes, @raphaelm is right. Additionally, the listening port will change.

We can actually make this work with previous installs, by running the service in the container as root, but dropping privileges in gunicorn (-u / -g). I personally don't like that approach very much, but it will maintain compatibility with previous installs, including keeping the bind on port 80.

I'll do a few tests, and update the pull request.

morisson:
Fair enough then :)

morisson:
I've just created a repo for a complete "dockerized" pretalx install: [https://github.com/bsideslisbon/pretalx-docker](https://github.com/bsideslisbon/pretalx-docker)
It includes a patch with the changes I've made in this pull request for it to work cleanly.

I'll still be improving it a bit, but it works pretty cleanly out of the box. I did this since I had some trouble making it work just with the dockerfile provided, and this now sets up all the needed dependencies, mysql, redis, including an smtp sink server for testing.
We'll be using pretalx as the CFP system for bsideslisbon this year!

If you want to have this repository in the main pretalx account, go ahead :)

Cheers, and thanks for your excellent work with pretalx.

rixx:
Sorry for taking so long to respond – all my energy is currently focused on making pretalx 1.0 release ready, so I've kind of ignored this part.

Your work on that repository is great, thank you very much! Currently, I'd love to move the repository into the pretalx org (since you offered, it's very appreciated), and remove the docker documentation in favor of a pointer towards the repository. Does that sound good for you?

morisson:
Hi Rixx,

sure, fine by me!

rixx:
@morisson If you give me access to the repository, I can use the official GitHub migration path so that it forwards all links etc. If that's too much of a hassle I can just re-push everything, your call

morisson:
@rixx Sure! I'll be giving you access.

rixx:
Thank you very much! It seems I can't move the repo though, so I guess I'll just push it as a new repository in the evening. I'll look through the setup information and then just link the README in our docs instead of the whole docker section we have now. Really cool, thank you again!

PR #379, opened 2018-03-26
[orga] Avoid duplicate listing of submissions

raphaelm:
Submissions with two speakers show up in the submission list twice, but *only* if the the list is filtered (using any of the available filters). This happens because by searching in ``speakers__name__icontains``, Django creates SQL that SELECTs not only from the submission table, but makes a cross product with the speakers table.

In this case, a simple ``distinct()`` is luckily sufficient to fix it.

## How Has This Been Tested?
Manually.

## Checklist


rixx:
Thank you! :cookie:

Issue #388, opened 2018-03-30
CI should run tests against all supported databases


raphaelm:
## Expected Behavior
Travis CI should run the test suite against all officially supported databases.

## Current Behavior
Travis CI uses only SQLite.

## Possible Solution
Define a test matrix for Travis CI that uses multiple config files.

raphaelm:
Context: We just had a bug (#387) that is only reproducible on PostgreSQL.

PR #389, opened 2018-03-30
Make all tests run on PostgreSQL

raphaelm:
Parts of the test suite currently rely on database quirks like auto-incrementing-columns starting at 1 in every test, natural ordering, etc. These tests fail on PostgreSQL, where those assumptions are not true.

## How Has This Been Tested?
Manually by running the test suite with PostgreSQL. We should make this automatic with #388.

## Checklist


rixx:
Thank you! :cookie:

PR #406, opened 2018-04-01
[orga] do not allow deletion of last orga member

### LINKED ISSUE #395: Prevent removing the last organiser of an event
… because this is a really bad idea.

luto:
Fixes #395 

This should work and has a test, but please tell if it's sane enough :sweat_smile:


rixx:
Thanks!

Issue #419, opened 2018-04-16
Unable to send emails


Rafiot:
I reinstalled from scratch the 0.5.0 release and pretalx doesn't send email any more.

## Expected Behavior

Send out a mail when a new reviewer is added.

## Current Behavior

Add a new reviewer, save, no email is sent.

## Steps to Reproduce (for bugs)

1. `https://<hostname>/orga/event/<conference>/settings/mail` -> configure and save (works, I see it in the logs)
2. `https://<hostname>/orga/event/<conference>/settings/team` -> add an email address -> nothing happen in the logs

## Context

Creating a team

## Your Environment

* Version used: v0.5.0

rixx:
Have you ticked the "use custom SMTP server" box? Have you tried this with and without active celery workers? As we run pretalx-0.5 in production, I can assure you that it does send emails.

PR #420, opened 2018-04-16
[doc] Add commands to check the logs

Rafiot:
Is "solves" #419...

/me slap myself hard.


rixx:
Haha, don't worry, it happens to everyone. Thank you for helping others not make the same mistake. (Or, you know, feeling even worse because there were docs …)

PR #426, opened 2018-04-30
Remove public domains as placeholders in user facing code

nazco:
This PR changes two user facing placeholders which use domains that are registered by other companies and so shouldn't be used as examples. To solve the problem I changed them to example.org like on other parts of the code.

## How Has This Been Tested?
Not needed.


rixx:
Not sure this is easier for people to understand, but if you insist …

PR #436, opened 2018-05-27
Extend API with submission image, speaker avatar and fix speaker biography

sreimers:
Extend API with submission image, speaker avatar and fix speaker biography

## Checklist

### DIFF
--- a/src/pretalx/api/serializers/speaker.py
+++ b/src/pretalx/api/serializers/speaker.py
@@ -1,5 +1,5 @@
 from rest_framework.serializers import (
-    CharField, ModelSerializer, SerializerMethodField,
+    CharField, ImageField, ModelSerializer, SerializerMethodField,
 )
 
 from pretalx.api.serializers.question import AnswerSerializer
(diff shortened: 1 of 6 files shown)


src/pretalx/api/serializers/speaker.py
```diff
@@ -12,19 +12,20 @@ class SubmitterSerializer(ModelSerializer):
 
     def get_biography(self, obj):
         if self.context.get('request') and self.context['request'].event:
-            getattr(obj.profiles.filter(event=self.context['request'].event).first(), 'biography', '')
+            return getattr(obj.profiles.filter(event=self.context['request'].event).first(), 'biography', '')
```
rixx: Good catch!

rixx:
Meh, GitHub reviews are weird.

What I meant to say was this: I'm not sure how having an "avatar.png" is helpful – we should return the full URL to the image, I think, so that API consumers can actually use the information provided.

sreimers:
Yes, the api already provides the full image/avatar url. a better documention hint would be maybe "https://example.net/media/avatar.png"

rixx:
The API does that if you use the `speakers/` endpoint, but for the `speakers` field inside the `submissions/` endpoint, it just shows "avatar.png". Could you have a look at that?

sreimers:
Ah, sure i review this.

sreimers:
ok done

rixx:
Thank you very much!

Issue #128, opened 2017-09-12
Make the agenda available offline with ServiceWorker


rixx:
Event networks are often flaky, and having a schedule (with a visible warning, naturally) available offline would be wonderful.

PR #449, opened 2018-06-18
[agenda] make schedule available offline with service worker

rahmonov:
This PR closes/references issue #128 . It does so by using service worker with the scope `/<event_name>/schedule/`. 

## How Has This Been Tested?
1. Start the development server.
2. Go to the event schedule page.
3. Switch off the server.
4. Reload the page and see that the schedule page is loaded with no problem even if the server is not started.

Not sure if I put `sw.js` in the right place...

### DIFF
--- a/src/pretalx/agenda/templates/agenda/schedule.html
+++ b/src/pretalx/agenda/templates/agenda/schedule.html
@@ -110,4 +110,12 @@
         </div>
     {% endfor %}
 </div>
+
+
+<script>
+  if('serviceWorker' in navigator) {
+    var scope = '/{{ request.event.name }}/schedule/';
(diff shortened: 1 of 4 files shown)


src/pretalx/agenda/templates/agenda/schedule.html
```diff
@@ -110,4 +110,12 @@ <h3 id="{{ day.start.date|date:"c" }}">
         </div>
     {% endfor %}
 </div>
+
+
+<script>
```
luto: If at all possible, please do not use inline scripts, as it makes it harder to write a good CSP. I am aware that we've got a few already, though ;)

rixx:
Very cool, thank you! Javacript files generally go into `/static`, but don't worry about it, I'll adjust it after the merge.

PR #456, opened 2018-07-15
Update Dockerfile

matedealer:
Changed from mysqlclient to mariadb-client.

## How Has This Been Tested?

I tested it on my arch linux.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.6
 
 RUN apt-get update && apt-get install -y git gettext \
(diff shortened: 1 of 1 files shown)


rixx:
This Dockerfile is deprecated and not recommended for use. If you want to run pretalx with docker, consider using https://github.com/pretalx/pretalx-docker.

matedealer:
Thanks for the comment. It may be nice to link the project in the documentation.

rixx:
It's mentioned [here](https://docs.pretalx.org/en/latest/administrator/index.html), but I'll also add it to the sub-page on installation, and add a comment to the Dockerfile. Thank you for bringing it up!

PR #470, opened 2018-08-05
Review Dashboard: Clarify German wording

raphaelm:
Currently, the review dashboard shows things like

```
9
Reviewer insgesamt
(13 insgesamt)
```

which doesn't give a way what's the difference between the two numbers.


rixx:
Thanks! :de: :gb:

PR #499, opened 2018-09-11
fix crash on go-live page, caused by i18n strings not having a len()

luto:
The go-live page in master crashes when I open it:

```
TypeError at /orga/event/evt/live
object of type 'LazyI18nString' has no len()

File "/home/luto/c3w/pretalx/src/pretalx/orga/views/event.py" in get_context_data
  94.         if not self.request.event.cfp.text or len(self.request.event.cfp.text) < 50:
```

I am not sure if that page is final and open to contributions yet, so accept or reject this MR at will :)

## How Has This Been Tested?

manually.

## Checklist

### DIFF
--- a/src/pretalx/orga/views/event.py
+++ b/src/pretalx/orga/views/event.py
@@ -91,9 +91,9 @@ class EventLive(EventSettingsPermission, TemplateView):
         context = super().get_context_data(**kwargs)
         warnings = []
         suggestions = []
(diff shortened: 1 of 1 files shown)


rixx:
Please just cast to ``str()`` instead – we'll check for missing translations at another time, for now it's only important that *something* will be shown on those pages.

PR #514, opened 2018-10-01
Added person id to file path

hnzlmnn:
If no person/submission identifier is used in the path of a file, multiple submissions to an event does not allow for equally named files. This could become a problem e.g. if submissions require something like a verification file (that could easily be named equally).

## How Has This Been Tested?
Just changed the file path. No tests were performed.

## Checklist

### DIFF
--- a/src/pretalx/submission/models/question.py
+++ b/src/pretalx/submission/models/question.py
@@ -9,7 +9,7 @@ from pretalx.common.urls import EventUrls
 
 
 def answer_file_path(instance, filename):
-    return f'{instance.question.event.slug}/question_uploads/{instance.person.code}/{filename}'
+    return f'{instance.question.event.slug}/question_uploads/{instance.submission.code}/{filename}'
 
 
 class QuestionManager(models.Manager):
--- a/src/pretalx/submission/models/question.py
+++ b/src/pretalx/submission/models/question.py
@@ -9,7 +9,7 @@ from pretalx.common.urls import EventUrls
 
 
 def answer_file_path(instance, filename):
-    return f'{instance.question.event.slug}/question_uploads/{filename}'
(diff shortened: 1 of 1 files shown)


rixx:
Have you had an actual issue with this? How did you encounter it? Django's file management should take care of this.

hnzlmnn:
Yes, we had an issue with this. We added a FileUpload question to a CfP. First, the link to the file does not show up in the question's answer. Only the filename is shown. To view submitted answers we had to go to `/orga/event/{event}/cfp/questions/{question}`.

When testing the submission the files were uploaded to `{event.slug}/question_uploads/{filename}`. When a second submission including an equally named file was sent, the answer overview showed that on file was uploaded correctly but the second only showed the filename and `(File missing)`.

A better approach to fix this would be renaming files to some hash/uuid and providing a download route that sets the original filename from the DB entry.

hnzlmnn:
Yeah, I realized that. If I have some time I'll push another commit that includes a none-check for submission to the path function. If the answer contains a submission, the path should reflect that. If it's a speaker/reviewer answer it should include a reference to the user which should be included into the path.

Is it possible that an answer doesn't have a user nor a submission reference?

If so, I'd prefer implementing the uuid4 approach.

rixx:
An answer should always have a user or a submission **so far**, but that's not guaranteed to stay that way. A person can also answer several questions, and a submission may also involve several questions.

hnzlmnn:
This could be addressed by adding the question id/slug to to path I guess?

To keep the option of having answers without user/submission, using an uuid4 for the filename would be preferred.

What's your view on this?

rixx:
Django, by default, just adds a short hash after filenames if the filename was already in use. I think this would be the best way forward.

luto:
There doesn't seem to be a lot happening here, feel free to open an issue or a new PR! thanks :)

Issue #531, opened 2018-11-11
Tests not running with different Tox config file


Nakaner:
## Expected Behavior

Tests should run even if Tox gets a different configuration file than `tox.ini` using the `-c FILENAME` command line option.

## Current Behavior

```
[michael@michael-probook pretalx]$ tox -vvv -e tests-sqlite -c tox-archlaptop.ini
  removing /home/michael/Entwicklung/github.com/pretalx/pretalx/.tox/log
using tox.ini: /home/michael/Entwicklung/github.com/pretalx/pretalx/tox-archlaptop.ini
using tox-3.5.3 from /usr/lib/python3.7/site-packages/tox/__init__.py
skipping sdist step
tests-sqlite start: getenv /home/michael/Entwicklung/github.com/pretalx/pretalx/.env
tests-sqlite reusing: /home/michael/Entwicklung/github.com/pretalx/pretalx/.env
tests-sqlite finish: getenv after 0.06 seconds
tests-sqlite start: developpkg /home/michael/Entwicklung/github.com/pretalx/pretalx/src
  /home/michael/Entwicklung/github.com/pretalx/pretalx/src$ /home/michael/Entwicklung/github.com/pretalx/pretalx/.env/bin/python /home/michael/Entwicklung/github.com/pretalx/pretalx/src/setup.py --name 
  /home/michael/Entwicklung/github.com/pretalx/pretalx$ /home/michael/Entwicklung/github.com/pretalx/pretalx/.env/bin/python -c 'import sys;  import json; print(json.dumps(sys.path))' 
tests-sqlite develop-inst-noop: /home/michael/Entwicklung/github.com/pretalx/pretalx/src
tests-sqlite finish: developpkg after 0.25 seconds
tests-sqlite start: envreport 
setting PATH=/home/michael/Entwicklung/github.com/pretalx/pretalx/.env/bin:/home/michael/.gem/ruby/2.3.0/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl
  /home/michael/Entwicklung/github.com/pretalx/pretalx$ /home/michael/Entwicklung/github.com/pretalx/pretalx/.env/bin/python -m pip freeze >/home/michael/Entwicklung/github.com/pretalx/pretalx/.env/log/tests-sqlite-7.log
tests-sqlite finish: envreport after 0.36 seconds
[...]

PR #532, opened 2018-11-12
Add field orga_notes to submissions for internal notes by the organisers

Nakaner:
As a Django noob, I am happy to be able to submit my first pull request.

I would like to share notes about the submission, e.g. things I discussed with the submitter, with other reviewers/organisers. I don't or can't write everything into the notes field because it is visible to the submitter.

notes-orga

## Checklist

- [ ] All new and existing tests passed. – unable to run them, see #531 


src/pretalx/orga/templates/orga/submission/content.html
```diff
@@ -97,6 +97,7 @@
         {% if form.abstract %}{% bootstrap_field form.abstract layout='event' %}{% endif %}
         {% if form.description %}{% bootstrap_field form.description layout='event' %}{% endif %}
         {% if form.notes %}{% bootstrap_field form.notes layout='event' %}{% endif %}
+        {% if form.orga_notes %}{% bootstrap_field form.orga_notes layout='event' %}{% endif %}
```
rixx: Why wouldn't the field be present? Do you want it to be hidden towards reviewers?

src/pretalx/submission/models/submission.py
```diff
@@ -124,6 +124,14 @@ class Submission(LogMixin, models.Model):
     notes = models.TextField(
         null=True, blank=True, verbose_name=_('Notes for the organiser')
     )
+    orga_notes = models.TextField(
```
rixx: Please call the field internal_notes instead.

rixx:
Just a short ping: are you still interested in pursuing this PR?

src/pretalx/submission/models/submission.py
Nakaner: The `notes` field is currently not exposed by the API as well. Shall I add it?

src/pretalx/submission/models/submission.py
```diff
@@ -124,6 +124,14 @@ class Submission(LogMixin, models.Model):
     notes = models.TextField(
         null=True, blank=True, verbose_name=_('Notes for the organiser')
     )
+    internal_notes = models.TextField(
+        null=True,
+        blank=True,
+        verbose_name=_('Notes by the organiser'),
+        help_text=_('Notes by the organiser for themselves (not visible for the speaker and the public)')
```
rixx: Please add a '.' at the end of the sentence, as the follow-up remark on markdown looks weird otherwise.

src/pretalx/submission/models/submission.py
```diff
@@ -124,6 +124,14 @@ class Submission(LogMixin, models.Model):
     notes = models.TextField(
         null=True, blank=True, verbose_name=_('Notes for the organiser')
     )
+    internal_notes = models.TextField(
+        null=True,
+        blank=True,
+        verbose_name=_('Notes by the organiser'),
```
rixx: I think calling the field 'Internal notes' would be better – otherwise you'll have "notes for the organiser" and "notes by the organiser" directly next to each other, and that would be more confusing than helpful.

Nakaner:
I rebased on current master and merged migrations (`manage.py makemigrations --merge`).

I decided against adding the `internal_notes` field to the API output in this pull request because it is not just adding a single line. Currently, the `/submissions` endpoint returns public fields only. Therefore, the view should be made aware about the permissions of the user.

src/pretalx/submission/migrations/0034_merge_20190304_0958.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 2.1.7 on 2019-03-04 09:58
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
rixx: Just a heads-up, I'll probably remove your migrations before the merge, since there's no real reason to keep merge migrations.

PR #539, opened 2018-12-01
Add GitHub authentication

vmx:
It's now possible to login with GitHub. The user won't need to create
a local account with username and password.

I tried to make this change as minimal and clean as possible. Though it's surely not ready to be merged, but I think it's good enough to get feedback on what is needed in order to get it merged.

## How Has This Been Tested?

Manually with logging in with a GitHub account. Also deleting the account again and logging in again. And that a lot of times during the development.

## Checklist

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -3,6 +3,7 @@
 Release Notes
 =============
 
+- :feature:`-` Login via GitHub is now possible.
 - :support:`-` A new pretalx plugin adds media.ccc.de as a recording provider – this plugin replaces the previously inbuilt capacity of pretalx to provide recording iframes. (This functionality was never directly exposed and only accessible via the pretalx shell. It is now deprecated and will be removed in a later version.)
 - :feature:`-` Plugins can now provide recording iframes (via the new ``register_recording_provider`` signal and other mixins).
 - :feature:`-` The new ``nav_event_settings`` plugin signal allows plugins to integrate their own settings pages next to the pretalx core pages.
--- a/src/pretalx/cfp/templates/cfp/event/login.html
+++ b/src/pretalx/cfp/templates/cfp/event/login.html
@@ -1,6 +1,8 @@
 {% extends "cfp/event/base.html" %}
 {% load bootstrap4 %}
+{% load cfp %}
 {% load i18n %}
+{% load socialaccount %}
 
 {% block title %}{% trans "Login" %} :: {% endblock %}
 
@@ -34,6 +36,10 @@
                 <a class="btn btn-block btn-link" href="{{ request.event.urls.reset }}">
                     {% trans "Reset password" %}
                 </a>
+                {% is_auth_provider_setup "github" as is_github_setup %}
+                {% if is_github_setup %}
+                <p><a href="{% provider_login_url 'github' %}{% if not request.GET.next %}?next={{ request.event.urls.user_submissions }}{% endif %}">Log In with GitHub</a></p>
+                {% endif %}
             </div>
             <div class="col-md-6">
                 <h4 class="text-center">{% trans "I need a new account" %}</h4>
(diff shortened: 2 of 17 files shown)


vmx:
It's great to see that you like the change. It was really meant as a starting point only to get third party authentication into the system eventually.

> adding one specific OAuth provider to pretalx seems very incongruent and random to me.

The plan surely is to have more providers supported. For keeping things simple, I started with one.
`django-allauth` uses the admin interface (which isn't enabled for pretalx) to setup other providers. It would also be possible to mimic that UI, so you could add any provider `django-allauth` supports. The advantage of having a custom interface for each "supported" provider is that you can leave out fields you don't need for that specific provider.

> make plugins like `pretalx-auth-github` or `pretalx-auth-twitter` etc a reality.

As those would all be handled via `django-allauth`, I'd rather do a `pretalx-auth-allauth` plugin instead which supports all providers. Or alternatively having `django-allauth` in the core and then having `pretalx-allauth-github` and `pretalx-allauth-twitter` plugins which just provide a nice customized UI for them.

>     * How do we handle /orga logins? They are not event specific, at the moment, while oauth configuration is per-event in the current draft.

To be honest, I didn't fully understand the distinction between `/orga` and the event specific stuff. It looked like use accounts are still central/unique to the whole system. So I just went ahead and made it work somehow. I think ideally you would configure one for the whole system and then enable different providers on an event basis.

rixx:
> The advantage of having a custom interface for each "supported" provider is that you can leave out fields you don't need for that specific provider.

I'm all for custom-built interfaces, hence the idea of offering different plugins for the task.

> > make plugins like `pretalx-auth-github` or `pretalx-auth-twitter` etc a reality.
> 
> As those would all be handled via `django-allauth`, I'd rather do a `pretalx-auth-allauth` plugin instead which supports all providers. Or alternatively having `django-allauth` in the core and then having `pretalx-allauth-github` and `pretalx-allauth-twitter` plugins which just provide a nice customized UI for them.

Yes, these are the two reasonable options, and I think having one `pretalx-oauth` plugin would be easier to maintain, and it would house common code parts, so we should go with that.

> > * How do we handle /orga logins? They are not event specific, at the moment, while oauth configuration is per-event in the current draft.
> 
> To be honest, I didn't fully understand the distinction between `/orga` and the event specific stuff. It looked like use accounts are still central/unique to the whole system. So I just went ahead and made it work somehow. I think ideally you would configure one for the whole system and then enable different providers on an event basis.

Well, generally things work on a per-event basis in pretalx – events that have nothing in common can share an instance. So at least the choice of which providers to offer on login would be event specific, but the /orga login page is not namespaced under an event (one of the reasons being that you might not even have an event yet, and would still want to log in).

vmx:
Thanks for the answers. What would the next steps be?

vmx:
I should rephrase it to: What should I work on in order to make GitHub authentication in Pretalx possible :)

rixx:
Sorry, yes, these last days were filled with stress and sickness, so I didn't get round to answering.

My proposal would be along these lines: I open a pretalx/pretalx-socialauth repository that will be the plugin for pretalx socialauth (of all kinds), with the goal of having GitHub at first (others as requested/required). You could work on that repository and figure out which plugin signals would be required in pretalx. I'd try to provide the authenticator via one class, as we do with [recording providers](https://docs.pretalx.org/en/latest/developer/plugins/recording.html) and [exporters](https://docs.pretalx.org/en/latest/developer/plugins/exporter.html).

Would you be willing to work on things like that? Does that sound like a feasible plan?

vmx:
That sounds great. I don't know much about signals yet, but I guess with the help of the examples above I can work something out.

rixx:
Sure, and I'll be happy to help – but in the end, you can look at the existing examples and either make a similar PR to pretalx, or just tell me about the interface you'd need, and I'll add them for you.

I added the [repository](https://github.com/pretalx/pretalx-socialauth/), put in the files generated by our plugin cookiecutter, and gave you access to it. Feel free to ping me with anything else you need!

PR #546, opened 2018-12-05
fix sentence negation

MaZderMind:
quick fix to german translation


rixx:
Oops, this is where automatic translation regeneration can bite you. Thanks!

PR #556, opened 2018-12-20
NorthSec is using pretalkx

h3xstream:
In case, you want to include an additional event using the platform.


rixx:
Thank you! Would you mind if I used your logo on the pretalx website among the users of pretalx?

h3xstream:
@rixx Yup It's ok.

PR #582, opened 2019-02-04
[WIP] Make questions related to a single SubmissionType only

Nakaner:
I have a use case where some questions are relevant to a single submission type only because our conference has both an academic and a community track.

This PR has a breaking change because it introduces a new ForeignKey relation from Question to SubmissionType, the migration might have unexpected side effects for events with a running CfP because a question now applies to a single submission type only if it applies to submissions.

I am no Django expert. Therefore, there might be better ways to model this. That's why I did not write any unit tests so far and would like to wait until the design in general is accepted.

invalid_combination
orga_events_sotm2018_cfp_questions
orga_events_sotm2018_cfp_questions_3_edit

## Checklist


Nakaner:
Existing unit tests will fail for this PR at the moment because they have been failing on the master branch as well for a few days.

rixx:
Thank you for your contribution! I'm not sure this is a good idea as it stands, though, for several reasons:

- Limiting question to submission types should rather be a multi-select, where a question can apply to one or several submission types, to avoid organizers having to duplicate questions for similar submissiont types.
- This should not be a breaking change – if no submission types are selected, the question should apply to all submission types, so that nothing changes with the upgrade.
- When creating a question that is not per-submission, the submission type should not be selectable.
- I don't think the submission type should be shown in the table – cramming all information in there makes it less usable.

Let me know if you have thoughts regarding this or if you need additional pointers!

Nakaner:
> * This should not be a breaking change – if no submission types are selected, the question should apply to all submission types, so that nothing changes with the upgrade.

My initial draft used a `ManyToManyRelation`. I wondered how to set the default to "reference all submission types of this event". I used the null value as replacement for "applies to all submission types" in order to make a question apply to all types even if a new type has been added. But this approach did not look good to me. Do you have better suggestions?

> * When creating a question that is not per-submission, the submission type should not be selectable.

I should hide the submission_type field in the form using JavaScript if target is not submission, shouldn't I?

> * I don't think the submission type should be shown in the table – cramming all information in there makes it less usable.

If one question applies to one submission type only, it is useful to be able to distinguish equal questions for different submissison types. But I agree that it is not necessary if a question can apply to any number of submission types.

rixx:
> I used the null value as replacement for "applies to all submission types" in order to make a question apply to all types even if a new type has been added. But this approach did not look good to me.

That's exactly what I would suggest.

> I should hide the submission_type field in the form using JavaScript if target is not submission, shouldn't I?

That would be best, yes! We do already have js in place to hide/unhide the choices editing parts, and warnings regarding required boolean questions, so that shouldn't be too bad to add.

Nakaner:
The breaking changes are now gone and replace by the NULL = all default you mentioned.

src/pretalx/cfp/views/wizard.py
```diff
@@ -55,7 +56,14 @@ def get(request, *args, **kwargs):
 
 
 def show_questions_page(wizard):
-    return wizard.request.event.questions.all().exists()
+    submission_type = wizard.get_cleaned_data_for_step('info')['submission_type']
+    return wizard.request.event.questions.exclude(
+        Q(target=QuestionTarget.SUBMISSION)
+        & (
+            ~Q(submission_type__name__contains=submission_type.name)
```
rixx: Please do not match on the name of the submission type, but on the type itself.

src/pretalx/event/migrations/0018_auto_20190207_1415.py
```diff
@@ -0,0 +1,18 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('event', '0017_auto_20180922_0511'),
+    ]
+
+    operations = [
```
rixx: Please remove this migration, it has no connection to this PR.

src/pretalx/submission/models/question.py
```diff
@@ -75,6 +75,12 @@ class Question(LogMixin, models.Model):
         verbose_name=_('question type'),
         help_text=_('Do you require an answer from every speaker or for every talk?'),
     )
+    submission_type = models.ManyToManyField(
+        to='submission.SubmissionType',
+        related_name='questions',
+        help_text=_('Is this question limited to some submission types only?'),
```
rixx: You're missing a `verbose_name` here. Also, it would be worth mentioning that leaving this field empty means including all submission types.

src/pretalx/submission/forms/question.py
```diff
@@ -31,8 +33,11 @@ def __init__(self, *args, **kwargs):
         self.queryset = Question.all_objects.filter(event=self.event, active=True)
         if self.target_type:
             self.queryset = self.queryset.filter(target=self.target_type)
+            if self.submission_type:
+                self.queryset = self.queryset.filter(Q(submission_type=self.submission_type) | Q(submission_type=None))
         else:
             self.queryset = self.queryset.exclude(target=QuestionTarget.REVIEWER)
+            self.queryset = self.queryset.filter(Q(submission_type=self.submission_type) | Q(submission_type=None))
```
rixx: This should only be executed if `self.submission_type` is set and the target_type is submission, since we might not be dealing with a submission question otherwise.

src/pretalx/submission/forms/question.py
```diff
@@ -31,8 +33,11 @@ def __init__(self, *args, **kwargs):
         self.queryset = Question.all_objects.filter(event=self.event, active=True)
         if self.target_type:
             self.queryset = self.queryset.filter(target=self.target_type)
+            if self.submission_type:
```
rixx: This should only happen if the `target_type` is submission, to be sure we do the correct thing here.

src/pretalx/person/migrations/0021_auto_20190207_1415.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.5 on 2019-02-07 14:15
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('person', '0020_auto_20180922_0511'),
```
rixx: Please remove this migration, since it has no bearing on this PR.

rixx:
In pretalx, questions can now be restricted to tracks. I have no *strong* feelings if we should allow to distinguish by submission type, too, what are your thoughts?

I'm slightly leaning on the "no" side, since we'd need to determine what happens if a question is restricted by track and by submission type: Is it an AND? Is it an OR? Do we forbid doing both? Etc.

Nakaner:
Abandoning this PR because a similar feature was introduced by commit 173a5908b49c8887a7fb34d593aea65d0a760640

rixx:
Let me take this opportunity and thank you for the work in this PR – as you probably saw, I reused parts of it when introducing the same feature for tracks, so your work found its way into pretalx (and saved me some time and effort) after all!

PR #604, opened 2019-03-02
[orga] refactor average_score away

vmx:
The submission model has a property called `average_score`, but the queryset
for submissions also returns the average score with a property called `avg_score`.

Unify those two into a single `avg_score` property.

## How Has This Been Tested?

Checked on a local instance if:

 - the average score is still displayed in the reviews dashboard
 - the average score is displayed in the "content" page of a submssion, if there there are any scores

## Checklist

### DIFF
--- a/src/pretalx/orga/templates/orga/submission/content.html
+++ b/src/pretalx/orga/templates/orga/submission/content.html
@@ -80,7 +80,7 @@
                     });
                     });</script>
         {% endif %}
-        {% if form.instance.average_score != None %}
+        {% if submission.reviews.count %}
         <div class="form-group row">
             <label class="col-md-3 col-form-label">
                 {% trans "Review score" %}
--- a/src/pretalx/orga/templatetags/review_score.py
+++ b/src/pretalx/orga/templatetags/review_score.py
@@ -35,7 +35,7 @@ def _review_score_override(positive_overrides, negative_overrides):
 
 @register.simple_tag(takes_context=True)
 def review_score(context, submission):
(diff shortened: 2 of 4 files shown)


rixx:
I'm not sure what you're trying to achieve. The `average_score` property allows easy access to a submission's average score. The fact that we add this information in one other place to one queryset doesn't really have a bearing on this – the `average_score` property handles the case of "I have one submission and want to see its average score".

vmx:
> I'm not sure what you're trying to achieve.

It's part of a bigger thing I'd like to change. I'd like to have a setting, so that a user can only see their own score and not the one from others. And that's easier if the data comes from a single place only. I would then also rename `avg_score` to `score` as it would display either the average score, or just the score the user rated.

> As you can see, one issue with this pull request is that the `review_score` tag can now be used only in one very specific place, and we won't be able to use it on other pages, e.g. simple submission pages.

Indeed. But currently it is only used at this particular place, hence I thought it made sense to remove duplicated functionality.

rixx:
The change you propose doesn't make any sense, even taking into account your higher goal: As things stand now, if I get a submission from the database, I won't be able to get its average score without re-typing the long annotation that goes into this process. Since this is a feature I use, I'm disinclined to let you remove it for no current reason.

> I'd like to have a setting, so that a user can only see their own score and not the one from others.

This is something that should be managed via a permission, then, as we handle all access-layer settings. In any case I think it would make much more sense to discuss this in the context of a larger PR where we could have a look at alternatives.

Issue #611, opened 2019-03-04
Reviewers can see current talk score in anonymous mode


glasnt:
## Current Behavior

When navigating to `event/<eventname>/reviews`, if a submission has at least one review, the current score is revealed. 

## Expected Behavior

A reviewer should not be able to see comments or review scores for submissions that have not been reviewed.

## Steps to Reproduce

1. Submit a talk into the system
1. Review a talk at least once
1. Create a new reviewer account
1. Access the `event/<eventname>/reviews` page
1. The current score is visible

## Context

Ideally, a new reviewer coming into review submissions should not see the reviewers of any other reviewer. This is currently available given that individual review comments and rankings cannot be seen on any one submission before review, but the aggregate score is still visible on the submission list on the review screen. 

## Your Environment

Testing project on pretalx.com, v0.9.0

rixx:
That's a very good point! We'd also need to take away sorting by review score for reviewers, since that would show ranking information implicitly.

vmx:
> We'd also need to take away sorting by review score for reviewers

I'd like to keep that feature, but instead of showing the average score, show the score you have given.

That's useful if you want to adjust your score later on. When you start reviewing, you don't have much context of the other talks. So you find a talk really good and then you discover that there are many talks that are even better. If you now can sort by score you can get an overview on how you distributed your scores and if you might want to adjust that.

vmx:
@glasnt To clarify that. Do you want to see the score of other reviewers once you've reviewed it yourself?

I am asking as it seems that I like a similar feature, but I certainly *don't* want to see the scores/comments from other reviewers at any point.

My workflow would be:

 - Disable "show scores/comments from other reviewers"
 - Let the reviewers do their job
 - Once the review phase is done, enable "show scores/comments from other reviewers" again. Now reviewers can discuss the outcome

glasnt:
It would be ideal to have the score be your personal score. 

However, the current way pretalx works is: 
 - you don't see any reviews or comments
 - you review a talk
 - then you see all other reviews and comments for that talk. 

This has merit, but it also allows for folks to see other's scores and potentially change their own (again, this can be useful)

If the data that can be used to calculate the average score can be seen after the initial review, then I think the aggregate score for that talk should be visible. 

The mechanics of a "reveal all the things" might be too much, rather than just adjusting the displayed score in that aggregate view. 

[...]

PR #617, opened 2019-03-08
Review phases

rixx:
This PR closes/references issue #611, as detailed in the proposal in that issue. Todo:

- [x] Add review phase model
- [x] Migrate existing settings to review phase model
- [x] Create sensible review phase for exsting events
- [x] Create better review phases for new events
- [x] Update review phase via regular cronjob
- [x] Provide interface for review phase CRUD
- [x] Implement activation and ordering
- [x] Implement new review permissions in all pages


vmx:
> * Provide interface for review phase CRUD + ordering

What is the ordering about? I'm asking as I have a special request about the ordering, I didn't bring up yet as I thought it could be done later, but now it might make sense to bring it up.

I'd like to have a custom ordering per reviewer. So that the ordering is the same every time a reviewers looks through the submissions, but the ordering is different between different reviewers.

The reason for this is for conferences where not all reviewers review all submissions. This way the each reviewer can start at their top and if they do enough reviews, eventually everything will get reviewed. This might also lower bias, as you might score proposals differently at the beginning or the end of the list.

I know that this doesn't work well with paging/Django's ORM. You would need to get the full list and sort it manually. But I think it would be great having this as an option with a warning that it might slow down the system.

rixx:
> I'd like to have a custom ordering per reviewer. So that the ordering is the same every time a reviewers looks through the submissions, but the ordering is different between different reviewers.

This is a) out of scope for this MR, and b) not going to happen in pretalx any time soon. Stable random ordering is horrible to mix with pagination, and it wouldn't be worth the effort, I feel. Reviewers can go with the "save and next" functionality which gives them a random not-yet-reviewed submission, so we have random ordering, but don't have to deal with the combination of stable per-user randomness.

vmx:
> This is a) out of scope for this MR,
I agree it is out of scope I just wanted to bring it up in case it interferes with the current work (which is does not).

> Stable random ordering is horrible to mix with pagination, and it wouldn't be worth the effort, I feel.

Stable random sort works with pagination (at least in my tests), it's more of a performance/load issue.

> Reviewers can go with the "save and next" functionality which gives them a random not-yet-reviewed submission

Oh, wow. I thought "next" means "next in the list". That's great and solves my problem.

rixx:
> Stable random sort works with pagination (at least in my tests), it's more of a performance/load issue.

Oh, cool! So what method did you use to achieve stable random sorting? I'm not aware of any one that works with SQLite (and I'm not ready to drop SQLite support for pretalx).

vmx:
> Oh, cool! So what method did you use to achieve stable random sorting? I'm not aware of any one that works with SQLite (and I'm not ready to drop SQLite support for pretalx).

What I meant with performance/load issues is that you need to query the full table (without pagination). Convert that into an array. You then do the stable random sorting on it and return that result instead of the queryset. Django will take care of the pagination automatically.

rixx:
> What I meant with performance/load issues is that you need to query the full table (without pagination).

I'd judge this to be inacceptable in terms of performance – I'm using pagination precisely to reduce wait times and load. Once we've pulled everything from the db, we might as well stop using pagination at all. Some databases can do seeded random sorting, but as I said: to me, it's not at all worth the effort.

The only way of achieving this that I'd accept would be to provide stable random sorting only if pretalx runs with the appropriate databases. I'd probably merge a MR providing this, but I'll definitely not tackle it myself unless as sponsored development.

PR #626, opened 2019-03-08
[common] do not pass a ['list'] to mail server for reply-to

luto:
values passed to headers['reply-to'] are not processed in any way and
thus end up in the actual headers just as they are. in the end the
client ends up with a header like:
```
reply-to: ['orga@event.com']
```
which confuses it.

using the reply_to parameter of djangos mail class correctly joins the
python list into a parsable list:

```
Reply-To: orga@event.com, someone@else.com
```

## Checklist


rixx:
Thank you!

PR #638, opened 2019-03-25
[orga] Allow hiding of notes to organiser

vmx:
It's now possible to hide the notes to organiser in the review phases.
Those comments might uncover details that are meant for the organiser
and not for the reviewer.

## How Has This Been Tested?

Tested manually within the dev environment. Checked if a user with "is reviewer" permissions can see/cannot see the notes.

## Checklist

### DIFF
--- a/src/pretalx/event/forms.py
+++ b/src/pretalx/event/forms.py
@@ -242,6 +242,7 @@ class ReviewPhaseForm(I18nModelForm):
         fields = [
             'name', 'start', 'end',
             'can_review',
+            'can_see_notes_to_organiser',
             'can_see_speaker_names',
             'can_change_submission_state',
             'can_see_other_reviews',
--- a/src/pretalx/event/models/event.py
+++ b/src/pretalx/event/models/event.py
@@ -385,6 +385,7 @@ class Event(LogMixin, models.Model):
                 position=0,
                 can_review=True,
                 can_see_other_reviews='after_review',
+                can_see_notes_to_organiser=False,
                 can_see_speaker_names=True,
                 can_change_submission_state=False,
             )
@@ -395,6 +396,7 @@ class Event(LogMixin, models.Model):
                 position=1,
                 can_review=False,
                 can_see_other_reviews='always',
+                can_see_notes_to_organiser=True,
                 can_see_speaker_names=True,
                 can_change_submission_state=True,
             )
@@ -519,6 +521,7 @@ class Event(LogMixin, models.Model):
 
                 can_review=True,
                 can_see_other_reviews='after_review',
+                can_see_notes_to_organiser=False,
                 can_see_speaker_names=True,
                 can_change_submission_state=False,
             )
--- a/src/pretalx/orga/permissions.py
+++ b/src/pretalx/orga/permissions.py
@@ -86,6 +86,12 @@ def is_event_over(user, obj):
 
 
 @rules.predicate
+def can_view_notes_to_organiser(user, obj):
+    event = obj.event
+    return event.active_review_phase and event.active_review_phase.can_see_notes_to_organiser
+
+
+@rules.predicate
 def can_view_speaker_names(user, obj):
     event = obj.event
     return event.active_review_phase and event.active_review_phase.can_see_speaker_names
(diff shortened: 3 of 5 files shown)


vmx:
Why isn't the CI triggered?

rixx:
I don't think this is a good idea. The notes to the organizer are not special in regards to personal information – abstracts and descriptions can just as easy identify a speaker. The real way forward here (and what is most requested by event organizers) is an editing/review process, allowing organisers to censor submitted texts before they reach reviewers.

vmx:
In our case (the FOSS4G academic track) it would've been helpful to have this feature. There we actually had the case were the identity was just given due to the notes to the organiser.

For the general track, I expect a lot of submissions and we have a tight schedule. So the organiser might not have the time to go through all the submitted notes before the review starts. There this feature would be really helpful.

rixx:
As an organiser, I've seen loads of mentioning of the speaker's identity in talk descriptions, be it via references to prior talks, links to presentation uploads, or just stating of the name in the text, so I still think this setting is needlessly narrow. I don't want to end up wit a suite of "hide $field from reviewers" in the end.

vmx:
@rixx It doesn't need to end up with lots of "hide $field from reviewers". To me the important thing is that reviewers can see as little as possible, restricted to only the things that are really needed. To me that's title and description.

So perhaps the setting could be change to "show only title and description to reviewers". My guess is that this what most people want if they hide the speaker name.

Also the help text for the notes currently says "These notes are meant for the organiser and won't be made public.". I'd expect it to be for the organiser only, not for reviewers. My workflow would be that only organisers can see those notes and if they want also the reviews to see it, they can copy it over the internal notes.

@glasnt I'd like to loop you into this issue as you seem to have similar review requirements as I have.

vmx:
I forgot to add, that i also want to make it easy to understand for outsiders if I outline the review process, I can simply say it's a blind review (which is probably already enough, as I expect mostwhere"reviewers will only see the title and the description" and not "reviewers will only see title, description and ... and ... and".

Nakaner:
The notes for the organiser field does contain remarks on the content quite often which do not fit into the abstract. For example, at an open source conference, submitters explain why their code has not been released as free software (yet).

From my point of view, the term "organisers" includes reviewers. The wording of the help text should be improved.

Although I can understand the points made by @rixx, I think that a system like Pretalx should not tell its managers (the organisers of a conference) how to organise a conference but leave them the freedom to do it their way. Otherwise, they will either look for another system or are forced to run their own fork if they have the capacities to maintain and run it.

Nakaner:
An implementation of #648 could fix the issues described by @vmx as well but seems to require more work to me.

vmx:
> The notes for the organiser field does contain remarks on the content quite often which do not fit into the abstract. For example, at an open source conference, submitters explain why their code has not been released as free software (yet).

In my case we solve the issue of things that don't necessarily belong to the abstract by an additional custom field (question) which asks e.g. about links to the source code. It's a text field, so submitters can also point out reasons why something might not be released yet.

I would restrict the "notes to organiser" field really to the organiser. I would expect people to put in things like "I can only attend if I get a visa, it takes 4 weeks to apply". Things that a limited group of people should see (the organiser), but not necessarily a reviewer.

rixx:
> Although I can understand the points made by @rixx, I think that a system like Pretalx should not tell its managers (the organisers of a conference) how to organise a conference but leave them the freedom to do it their way. Otherwise, they will either look for another system or are forced to run their own fork if they have the capacities to maintain and run it.

While that's true, it's only true up to a point – if pretalx makes everything configurable, people will not want to engage with its huge jugle of settings. If pretalx makes nothing configurable, people will not want to use it since the workflow doesn't fit. I'm trying to find a balance here, and erring on the side of caution, since it's nearly impossible to remove settings once they have been introduced.

rixx:
I'd be more inclined to add a field of "visible_to_reviewers" to question answers, tbh.

vmx:
> I'm trying to find a balance here,

I fully agree with it. I've seen system that were unusable as everything could be configured.

Though I don't yet give up ;)

Especially for academic tracks, it's common that reviewers only see the title and the abstract. I'd like to be able to run those tracks with pretalx.

In cases cases where the notes to the organiser might contain valuable information, the workflow could be, that the organiser just pastes that into the internal notes, this way the reviewers could see it.

We did use pretalx for the academic track for the FOSS4G, people were putting their names into the "notes to organiser". I think the reason is that there things work differently from open source conferences. I would expect someone to submit the abstract, but for them it's also important to put the full list of authors (in specific order) somewhere, which aren't necessarily the speakers.

One could argue that you could make that a question, but I think you can't foresee all questions, so a default free-form filed like "notes to organiser" that only organisers can see would make sense to me.

vmx:
> I'd be more inclined to add a field of "visible_to_reviewers" to question answers, tbh.

After writing my previous comment I realised that this might solve my problem. I could de-activate the "notes to organiser" and just add a freeform field to the questions that reviewers won't see.

It would make the workflow as nice with the "notes to organiser" field, but that's an OK trade-off.

rixx:
I just added a feature that allows you to specify if reviewers are allowed to see specific questions. That should solve this issue, I think!

PR #643, opened 2019-03-30
[cfp] do not show submission types form ALL events in some cases, fix #642

### LINKED ISSUE #642: Having only one submission type shows all submission types
## Current Behavior
If an event only uses one submission type, the submission form shows submission types for ALL events

## Expected Behavior
Only submission types for the current conference should populate the dropdown. 
Added bonus: if only one, the field should be read-only/text. 

## Steps to Reproduce

1. https://pretalx.com/orga/event/[YOUREVENT]/cfp/types/
2. Enter only one type
3. Navigate to https://pretalx.com/[YOUREVENT]/submit/
4. Click the dropdown for "Submission type"

Screen Shot 2019-03-30 at 5 35 39 pm

### Work around

Enter more than one submission type. 

## Your Environment

Pretalx hosted environment

luto:
fixes #642

## How Has This Been Tested?

via the supplied tests and manually

## Checklist

### DIFF
--- a/src/pretalx/submission/forms/submission.py
+++ b/src/pretalx/submission/forms/submission.py
@@ -66,14 +66,12 @@ class InfoForm(RequestRequire, PublicContent, forms.ModelForm):
         pks = set(types.values_list('pk', flat=True))
         if instance and instance.pk:
             pks |= {instance.submission_type.pk}
+        self.fields['submission_type'].queryset = self.event.submission_types.filter(
+            pk__in=pks
+        )
         if len(pks) == 1:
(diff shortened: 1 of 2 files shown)


luto:
the current `disabled = True` method [does not work](https://stackoverflow.com/a/325038/2486196) because browsers won't send disabled fields back to the client at all. A similar cleanup is probably needed for `content_locale` as well, but I gotta run now.

rixx:
Hm. We use `disabled=True` in a couple of places, but mostly to actually make fields read-only. I'll remove the ``disabled=True`` from locales, but I don't think it'll make a difference in this case.

PR #651, opened 2019-04-07
Speed up tests by 50% by not compiling SCSS and caching templates

raphaelm:
When working on pretix, I found this speed-up opportunity and thought you might like it too :) On my machine, it gets the pretalx test suite down from around 100 seconds to around 45 seconds, which I think is quite impressive.

The two optimizations are:

* Cache templates in memory during a test run
* Do not compile SCSS during testing

The second is the major speed gain, but of course has the downside of, well, SCSS not being compiled. It would be easily possible to implement caching instead of just disabling it during tests, but I don't think its worth it, since any compilation errors in the SCSS codebase will be found by the ``rebuild`` command that is already run on CI.

## How Has This Been Tested?
With the test suite.
## Checklist


rixx:
Wow, yeah, thanks!

PR #657, opened 2019-04-14
Style

rixx:
Do we want this?


rixx:
Let's wait to see where DEP8 goes, and then look what we might want.

PR #659, opened 2019-04-16
[mail] Fix migration

vmx:
The migration might fail if an email address is not all lowercase.

The problem is that the address gets lowercasesed first and if there
is a user with this email address, then it is removed from the list
of addresses. But that removal expects the original address and not
the lowecases one.

## How Has This Been Tested?

I've run it locally on an instance where the migration was failing and now it passes.

## Checklist


vmx:
I can confirm that https://github.com/pretalx/pretalx/commit/c4fe92fe37ce1e433f9a3e1cf61c0567a3f88842 fixes the issue I wanted to fix with this PR.

PR #663, opened 2019-04-20
README: added Camp++

dnet:
The README states

> If you use pretalx for your event, please tell us about it!

We use it, so hereby we tell you about it :wink:


rixx:
Thank you! :hungary:

PR #664, opened 2019-04-20
[common] allow users/orga to upload SVGs

luto:
pretalx allowed SVGs as Event header image in the past, but currently doesn't. I'm not sure if this change was intentional, so you're now reading this PR instead of a commit in master :)

## How Has This Been Tested?

in production :)

## Checklist

### DIFF
--- a/src/pretalx/common/forms/fields.py
+++ b/src/pretalx/common/forms/fields.py
@@ -8,7 +8,7 @@ from pretalx.common.forms.widgets import (
     ClearableBasenameFileInput, PasswordConfirmationInput, PasswordStrengthInput,
 )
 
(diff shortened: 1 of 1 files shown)


luto:
site note: our event currently has a SVG image set, so we cannot edit any other event settings, because the form always fails because of the SVG :grin:

rixx:
SVG image uploads are not possible any more due to security concerns, no. I had hoped for the longest time to make some sort of SVG validation a part of pretalx, but as the problem has a LOT of complexity, and I haven't seen any convincing solutions, SVG uploads are not allowed any longer.

The only exception is if a user marked as `is_administrator` comes along, since they already are presumed to have root access and should know what they are doing.

Issue #680, opened 2019-04-26
Outgoing mails are not sent


kannes:
I receive no mails on new submissions (I did configure it to send them) or if I send a mail to all authors (including me).

I checked the network traffic with tshark:
- When using the SMTP testing on `/orga/event/myevent/settings/mail` I do see communication between my host and the remote smtp server I configured.
- But when other normal actions should trigger outgoing mail nothing happens.

There are no explicitly related errors in `/var/pretalx/data/logs/*`, just lots of
```
...
WARNING 2019-04-26 12:18:00,865 django.request log Not Found: /api/values
WARNING 2019-04-26 12:18:00,865 django.request log Not Found: /api/values
WARNING 2019-04-26 12:19:07,134 django.request log Not Found: /api/values
WARNING 2019-04-26 12:19:07,134 django.request log Not Found: /api/values
WARNING 2019-04-26 12:24:11,455 django.request log Not Found: /api/values
WARNING 2019-04-26 12:24:11,455 django.request log Not Found: /api/values
WARNING 2019-04-26 12:29:15,568 django.request log Not Found: /api/values
WARNING 2019-04-26 12:29:15,568 django.request log Not Found: /api/values
```

## Your Environment

Ubuntu 18-04, pretalx==0.9.0

kannes:
I just saw https://github.com/pretalx/pretalx/issues/419 and will look into that Celery stuff. I did not see that as being required during installation but maybe I overlooked some details...

kannes:
Yeah, that fixed it. I will update the docs to indicate that celery is indeed a requirement, not an option.

kannes:
On the upside, the previously unsent mails were all sent after I had the worker running. :)

rixx:
Celery is a requirement, but running separate celery workers isn't one, if you configure your instance appropriately. If you have a celery section in your settings, celery workers will have to run.

kannes:
So my mistake was "user configured `[celery]` in config but is not running workers". Should that result in a error or warning somewhere?

-----

Now I am confused...

I admit that I did *not* read https://pretalx.readthedocs.io/en/latest/administrator/configure.html#configure at all, I only followed https://pretalx.readthedocs.io/en/latest/administrator/installation.html

### Celery
So I could just leave out the `[celery]` part in `/etc/pretalx/pretalx.cfg` file and it would work without the workers?

https://pretalx.readthedocs.io/en/latest/administrator/configure.html#the-celery-section says: 

[...]

PR #681, opened 2019-04-26
docs: make sure the pretalx-worker is properly installed

kannes:
This makes using Celery not an optional "if you decide to" but, as the installation is otherwise not working out of the box, a requirement.

If this is actually optional and advanced users can use pretalx without, I would recommend splitting optional things into a separate document for those users.

This PR references issue https://github.com/pretalx/pretalx/issues/680

## How Has This Been Tested?
Not tested from scratch. Based on my installation on Ubuntu 18.04 today.

## Checklist

### DIFF
--- a/doc/administrator/installation.rst
+++ b/doc/administrator/installation.rst
@@ -154,7 +154,7 @@ named ``/etc/systemd/system/pretalx-web.service`` with the following content::
     [Install]
     WantedBy=multi-user.target
 
-For asynchronous execution of long-running tasks via Celery, such as outgoing mail),
+For asynchronous execution of long-running tasks via Celery (such as outgoing mail),
 you'll also need a second service ``/etc/systemd/system/pretalx-worker.service``
 with the following content::
 
--- a/doc/administrator/installation.rst
+++ b/doc/administrator/installation.rst
@@ -154,9 +154,9 @@ named ``/etc/systemd/system/pretalx-web.service`` with the following content::
     [Install]
     WantedBy=multi-user.target
 
-If you decide to use Celery (giving you asynchronous execution for long-running
-tasks), you'll also need a second service
-``/etc/systemd/system/pretalx-worker.service`` with the following content::
+For asynchronous execution of long-running tasks via Celery, such as outgoing mail),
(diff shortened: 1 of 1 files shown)


rixx:
This is incorrect – it's perfectly valid and possible to run pretalx without celery worker processes (indeed, on the current release it was necessary to do in combination with Python 3.7 for a long time due to Celery's rather slow release cycle). I have several pretalx instances running in production without celery worker processes, and they are definitely capable of sending emails.

kannes:
Then the existing docs would be wrong, they say (emphasis mine):

> If you decide to use Celery (giving you asynchronous execution for long-running tasks), you’ll also **need** a second service /etc/systemd/system/pretalx-worker.service with the following content:

rixx:
Yes, that is correct – if you have a `celery` section in your configuration, pretalx will assume you have a task runner out there, and will defer all tasks to a celery queue. If celery is not configured, pretalx will not depend on a task runner. Celery (the library) is installed in any case, as it is a dependency of pretalx, but if it is not *configured*, it will not behave as a task runner, and all potentially async tasks are executed directly (blocking, sync) instead.

kannes:
Then I proposed the following note at the end of https://docs.pretalx.org/en/latest/administrator/installation.html#step-4-configuration:

```
.. note:: The suggested configuration above includes a ``[celery]`` section.
          This means that you *do* use Celery if that comes up below.
```

Maybe with a link to https://pretalx.readthedocs.io/en/latest/administrator/configure.html#the-celery-section

No PR because I kept this all on GitHub for now and forks are icky once out-of-date.

rixx:
As of c30b6067 you can see the current celery queue length in the new admin panel, which might help debugging.

kannes:
That sounds like a great solution. I am super impressed by your work!

rixx:
Thank you! And also thank you for your continued feedback – I'm sorry the last interactions were mostly me shutting down your ideas, but I'm definitely happy to hear your experiences and to figure out ways to make them better. (Just gotta keep the overall project direction in mind at the same time.)

kannes:
Your replies were always clearly motivated by a strong vision how things *should* work, no need to feel sorry in any way!

PR #683, opened 2019-04-28
README: Add Klimacamp & Degrowth Summerschool

shoetten:
We migrated from frab and are very happy so far :)
Keep up the awesome work!


rixx:
Nice, thank you for letting us know!

PR #711, opened 2019-06-01
Remove placeholder address

glasnt:
Remove the placeholder address in the Email Settings view. 

**Why?**

Previously, if you had no value in this field, the help text would tell you:

> Sender address for outgoing emails. Leave empty to use the default address: noreply@pretalx.com

However, the placeholder for that input field defaults to be your event email. 

two fields

The contrast difference in the form and entered value means that it *looks* like you have set a value, but it's actually empty, using the default. (In this example, the first field has no value, only placeholder; the second field has a value)

By removing the placeholder, administrators won't confuse a value for a lack of value. Also, the helptext will be more accurate ("Leave empty" actually has an empty field, with no placeholder)

Mock up of result: 
Mock up of result


rixx:
Thank you! 🎀

PR #720, opened 2019-06-04
be more python3.x agnostic

minusf:
Most packages and `venv` symlink python3 to whatever they have installed. This will cover all 3.6+ versions.  The minimal python requirement is clear from the docs and `setup.py` checks for 3.6 so i don't see any risk here... It just makes the dev's life easier...

## Checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,7 @@ setupdir = src
 
 
 [testenv]
-basepython = python3.7
+basepython = python3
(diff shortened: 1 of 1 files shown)


rixx:
Ah, thank you – I don't use tox at all, so I never ran into this.

minusf:
but the development docs (ab)use tox everywhere :}  (i prefer makefiles)

PR #721, opened 2019-06-04
installation fails if configuration does not use sqlite

minusf:
If the configuration in **Step 4** uses a non-sqlite DB, **Step 5** `pip install --user -U pretalx` (before installing the DB driver) will fail. Swap the order in the documentation.

Also add a missing "with" in "pretalx works your choice of database backends".

(TBH i am not a fan of `pip install --user` as it doesn't scale if the user in question runs multiple apps (whether that's a good/bad is another discussion), but i still prefer vanilla venvs dedicated only to the app.)

## Checklist

### DIFF
--- a/doc/administrator/installation.rst
+++ b/doc/administrator/installation.rst
@@ -108,21 +108,14 @@ Check out :ref:`configure` for details on the available configuration options.
 Step 5: Installation
 --------------------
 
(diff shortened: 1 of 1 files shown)


rixx:
Thank you, much clearer for first-time installation.

PR #723, opened 2019-06-10
rewrite HTML export

luto:
Rewrite the HTML export! It show no longer uses django-bakery, but the django test client to make "real" WSGI requests. This makes it a little slower, but more closer to actual requests. It also gets rid of a lot of custom view code.

## How Has This Been Tested?

Manually and using the already present tests. It should work as a drop-in replacement.

N/A

## Checklist


src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+    """ find URLs of additional files needed to render `html`, e.g. images, js, css, ... """
+    soup = BeautifulSoup(html, "lxml")
+
+    for asset in soup.find_all(['script', 'img', 'link']):
+        if asset.name in ['script', 'img']:
+            yield asset.attrs['src']
+        elif asset.name == 'link':
+            if asset.attrs['rel'][0] in ['icon', 'stylesheet']:
+                yield asset.attrs['href']
+        else:
```
rixx: How do you arrive at this else, when we only iterate over `soup.find_all(['script', 'img', 'link'])`?

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+            raise NotImplementedError()
+
+
+def find_urls(css):
+    return re.findall(r'url\("?(/[^")]+)"?\)', css.decode('utf-8'), re.IGNORECASE)
+
+
+def urls_event_talks(event):
+    talks = event.submissions.filter(
+        pk__in=event.current_schedule.slots.all().values_list('pk', flat=True)
```
rixx: I'm confused about this line – what's te difference as opposed to `talks = event.current_schedule.slots`?

rixx: You might also just want to use `event.talks`, come to think of it.

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+        yield talk.urls.public
+        yield talk.urls.ical
+
+
+def urls_event_speakers(event):
+    from pretalx.person.models import SpeakerProfile
+
+    speakers = SpeakerProfile.objects.filter(
+        user__submissions__slots__schedule__published__isnull=False
+    ).select_related('user', 'event').prefetch_related('user__submissions').distinct()
```
rixx: Would `event.speakers` be a valid alternative here?

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+
+    for speaker in speakers:
+        yield speaker.urls.public
+        yield speaker.urls.talks_ical
+
+
+def urls_schedule_versions(event):
+    for schedule in event.schedules.all():
```
rixx: Or just use `event.schedules.filter(version__isnull=False)`?

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+
+
+def dump_content(dest, path, content):
+    if path.endswith('/'):
+        path = path + 'index.html'
+
+    path = os.path.join(dest, path.lstrip('/'))
+
+    try:
+        os.makedirs(os.path.dirname(path))
```
rixx: os.makedirs has an `exist_ok=True` flag. An alternative more in line with code in other places would be 

```python
from pathlib import Path

path = Path(dest) / path.lstrip('/')
path.parent.mkdir(parents=True, exist_ok=True)
```

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+                    css_assets |= set(find_urls(content))
+
+            logging.info(f"exporting {len(css_assets)} assets linked to in CSS")
+            for url in (get_path(urllib.parse.unquote(url)) for url in css_assets):
+                logging.debug(url)
+                content = get_mediastatic_content(url)
+                dump_content(dest, url, content)
+
+
+def rm_rf(path):
```
rixx: I think a more speaking name would be nicer. `delete_directory` or something like that?

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+    try:
```
rixx: Can I haz a `with suppress(FileNotFoundError)` here?

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+    except FileNotFoundError:
+        pass
+
+
+def get_export_path(event):
+    return settings.HTMLEXPORT_ROOT / event.slug
+
+
+def get_export_zip_path(event):
+    return str(get_export_path(event)) + '.zip'
```
rixx: ``get_export_path(event).with_suffix('.zip')``

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
-                    )
-                    output_dir = self.get_output_zip_path(event)
-        self.stdout.write(str(output_dir))
-
-    def build_views(self):
-        for view_str in self.view_list:
-            view = get_callable(view_str)
-            view(_exporting_event=self._exporting_event).build_method()
+        export_dir = get_export_path(event)
+        tmp_dir = str(export_dir) + '-new'
```
rixx:
```python
tmp_dir = export_dir.with_name(export_dir.name + '-new')
```

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+
+        rm_rf(tmp_dir)
+        os.mkdir(tmp_dir)
```
rixx:
```python
tmp_dir.mkdir()
```

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+
+        try:
+            export_event(event, tmp_dir)
+            rm_rf(export_dir)
+            os.rename(tmp_dir, export_dir)
```
rixx:
```
tmp_dir.rename(export_dir)
tmp_dir = export_dir
```

(Yeah, pathlib renaming is super weird.)

src/pretalx/agenda/management/commands/export_schedule_html.py
```diff
@@ -1,74 +1,230 @@
...
+        finally:
+            rm_rf(tmp_dir)
+
+        if options.get('zip'):
+            make_archive(
+                root_dir=settings.HTMLEXPORT_ROOT,
+                base_dir=event.slug,
+                base_name=os.path.splitext(get_export_zip_path(event))[0],
```
rixx: `get_export_zip_path(event).stem`

src/setup.py
```diff
@@ -116,10 +116,10 @@ def run(self):
         'vobject==0.9.*',  # 2017-06, http://eventable.github.io/vobject/ look for "release"
         'whitenoise==4.1.*',  # http://whitenoise.evans.io/en/stable/changelog.html
         'zxcvbn==4.4.*',  # Nothing? https://github.com/dwolfhub/zxcvbn-python/issues/38
+        'beautifulsoup4',
```
rixx: Dependencies are currently ordered alphabetically, pinned to a fixed version, and link their changelog for ease of upgrades. Could you please move bs4 up, pin its version and add a link?

src/pretalx/agenda/management/commands/export_schedule_html.py
luto: doesn't `tmp_dir = export_dir` remove `htmlexport/dj19` instead of `htmlexport/dj19-new` in the next line?

rixx: Yeah, ignore the assignment suggestion if you never access `tmp_dir` again.

Issue #648, opened 2019-04-02
Make help texts for hard-coded CfP fields editable


Nakaner:
## Problem you are facing

Some fields of CfP are hard-coded and can only be enabled or disabled. They are useful but their description can be confusing to submitters or lack necessary information. There are two ugly workarounds at the moment:

* Disable the hard coded field and replace it by a custom field. This has the disadvantage of not being integrated in various exports and the programme.
* If running Pretalx yourself, edit the template and maintain your own fork.

## Possible Solution

Make help text editable.

## Context

For example, in Pretalx, an "abstract" is a short description one or two sentences long. A "description" is a longer text about the submission. In the academic context, an abstract is multiple hundreds of words long.

At FOSSGIS, the short text is printed in a booklet and both texts are shown in the online programme and the app. As an organiser, I would like to be able to edit the help text to explain that.

This feature request is a result of our experiences at FOSSGIS 2019.

rixx:
I'd prefer to choose help texts that are no issue for anybody. Making all help texts for all fields in pretalx editable is quite a lot of settings, and settings fatigue in applications is a thing. I'd rather find a solution that works for everybody (and it may well be to remove the help texts entirely.

vmx:
> Making all help texts for all fields in pretalx editable is quite a lot of settings, and settings fatigue in applications is a thing.

Agreed.

An alternative idea, not sure if I like it myself. Could there be a pretalx plugin that allows to upload custom templates? This way power users could override help texts from within the templates, but wouldn't need to run it themselves (in case it's installed on the hosted instance).

Nakaner:
Custom templates also allow to add longer explanation texts. There could only be the issue of the loading content (images etc.) from third parties which is not coverd (but should mentioned) in the privacy policy of pretalx.com. In addition, one should check how safe it is to use Django templates submitted by users (or how powerful they are).

rixx:
User uploaded Django templates are absolutely never going to happen – Django templates have access to the complete database, and while they can't quite execute arbitrary code, they're very close to that.

Plugins should already be able to provide custom templates just by adding them to their templates folders – have you checked that?

rixx:
I've removed the help_text from the abstract and description fields, so that only the markdown note is left, and potential length restriction notifications. I'd say that any in-depth help in filling in the fields would be badly placed in our help text, since it's small and not great to read for longer texts, and should rather be placed elsewhere. Would you agree?

PR #768, opened 2019-09-04
CfP: allow to customise help texts of InfoForm

Nakaner:
Administrators of an event can now customise a few fields of the InfoForm of the submission wizard.

This pull request addresses the issue I described in #648. At previous events I had the issue that speakers did not fully understand what to enter in some fields of the submission form. Examples:

* Speakers should use complete sentences with a verb in the short description because it is inteded for printing
* Different conferences have a different opinion whether the selection of the right track is the responsibility of the speaker or not.
* The long description should not rely on the short description. (That's a decision of the programme committee depending how they present the long description.

## How Has This Been Tested?

Unit tests are currently failing due because the I18nTextFields are initialised with the return value of django.utils.translation.gettext_lazy which is obviously wrong. I will fix this. However, I would appreciate feedback on the general design.

## Checklist


rixx:
I'm honestly not sure that this is a good idea. Here is why I'm sceptical:

- It's a feature few organizers would use.
- It takes a lot of screen space, and as such contributes to settings fatigue.
- It adds further database access where I'd like to reduce it.
- Broken window: Once we allow to edit one front-end form, sooner or later edits to all of them will be requested (e.g. stopping at the Info form and not including the Person form will be actually confusing to users)

Nakaner:
This pull request addresses a feature request which arised during the post-event evaluation of FOSSGIS-Konferenz 2019. We had lots of speakers who were unsure how to fill the form. We assume that the difficulty to fill the form drives some speakers away, especially those whom we would like to support: new speakers, women, students, people without commercial relations to OSM/FOSSGIS. One could say that all the help could go into the call for papers. However, the call for papers is already long enough and I assume that people do not re-read the CfP text if they are unsure how to fill out the form.

> It takes a lot of screen space, and as such contributes to settings fatigue.

I can move the settings into a separate view if that would address this concern.

> It adds further database access where I'd like to reduce it.

I am not against moving these settings into a file. If we do so, all the conference I am responsible for will have to host Pretalx themselves if they want different help texts.

> Broken window: Once we allow to edit one front-end form, sooner or later edits to all of them will be requested (e.g. stopping at the Info form and not including the Person form will be actually confusing to users)

I thought adding another field for the biography field but I wanted to start with a smaller improvement first. The good news is that there are currently only few fields left speakers are asked to fill out and which do not have a customisable help text yet: title, biography, speaker name, email, password. All other fields are questions to be configured by the organiser. These questions have a "customisable" help text already.

rixx:
Hereby superseded by #809

Issue #766, opened 2019-08-30
Deal with changing time zones


rixx:
When organizers change the event timzone, we can either keep all talks at their configured time, which will move them in their time relative to the event time zone – or we can move all talks so that their relative time stays the same (e.g. 10am remains 10am).

Currently we do the first of these things, but I'm afraid the second would be more practical.

vmx:
I needed exactly that and wrote a script to do that. This might help someone working on that feature: https://gist.github.com/vmx/842af29b8e40814720be4799f1ccad36

insacuri:
I'll take a looksee

PR #781, opened 2019-09-17
Update times when timezone is changed for an event (WIP)

insacuri:
When the timezone is changed after a number of talks have been scheduled, the difference in timezone shifts the time it is scheduled for (e.g., a 10am talk will be moved to 11am in the schedule if changing from Europe/London to Europe/Paris time)

This PR closes/references issue https://github.com/pretalx/pretalx/issues/766. It does so by changing all the talk slots and availability times when the timezone is changed on the settings form.  It only attempts to do any checks/changes if the timezone has been changed so it won't impact performance.

## How Has This Been Tested?
Not yet, still a work in progress... (things that are currently broken are commented in the code)

## Checklist

### DIFF
--- a/src/pretalx/orga/views/event.py
+++ b/src/pretalx/orga/views/event.py
@@ -50,7 +50,7 @@ class EventSettingsPermission(EventPermissionRequired):
 
 
 class EventDetail(ActionFromUrl, EventSettingsPermission, UpdateView):
-    momdel = Event
+    model = Event
     form_class = EventForm
     permission_required = 'orga.change_settings'
     template_name = 'orga/settings/form.html'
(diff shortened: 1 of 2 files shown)


src/pretalx/orga/views/event.py
```diff
@@ -38,15 +39,18 @@
...
+logger = logging.getLogger(__name__)
+
 
 class EventSettingsPermission(EventPermissionRequired):
     permission_required = 'orga.change_settings'
 
 
 class EventDetail(ActionFromUrl, EventSettingsPermission, UpdateView):
-    model = Event
+    momdel = Event
```
rixx: Typo? 👀

insacuri: yup... ^^

PR #788, opened 2019-09-29
API Endpoints /submission and /talks: Return URL, not Internal Storage Path for Resources

Nakaner:
Use `FileField.url`, not `FileField.path`. See also https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.FileField.storage

While the frontend (`/orga/event/{event_slug}/submissions/{code}`) uses the `url` attribute, the API serialiser used `path` which returns the path on the disk. The unit tests did not point out this mistake because the compared the result against `path` as well.

## Checklist

- [x] ~I have updated the documentation accordingly.~ not necessary
- [x] All new and existing tests passed. (except those which fail on current master branch as well)

### DIFF
--- a/doc/api/resources/submissions.rst
+++ b/doc/api/resources/submissions.rst
@@ -31,7 +31,7 @@ image                                 string                     The submission 
 answers                               list                       The question answers given by the speakers. Available if the requesting user has organiser permissions.
 notes                                 string                     Notes the speaker left for the organisers. Available if the requesting user has organiser permissions.
 internal_notes                        string                     Notes the organisers left on the submission. Available if the requesting user has organiser permissions.
-resources                             object                     Files the speaker has uploaded for this submission. ``{"resource": "/path/to/file", "description": "Slides"}``
+resources                             object                     Files the speaker has uploaded for this submission. ``{"resource": "https://hostname.tld/path/to/file", "description": "Slides"}``
 ===================================== ========================== =======================================================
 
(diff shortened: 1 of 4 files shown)


rixx:
Oh, yes, absolutely. Thank you!

rixx:
No, this one was intentionally – the API does tell you the base URL, after all (or you know it already), but since some events use their own domains, prescribing a domain here feels like a bad idea – especially considering that there can be HTML exports involved.

Nakaner:
Well, the following response is not helpful for me:

```json
"resources":[{"resource":"/usr/share/webapps/pretalxprod/media/Skowron_is_your_osm_app_spying.pdf","description":"slides"}]
```

rixx:
Ah, I wasn't aware that this was the issue from your description, apologies. I would still maintain that the full URL would be a mistake in this case, though. I agree that `resource.url` would be the correct property to use, but I can't see where you saw that property yielding an absolute URL? You documented your change as using absolute URLs including the domain name …

tyrasd:
has this been deployed already? I still see internal paths in pretalx' api output…

rixx:
I assume you mean on pretalx.com? Irregular updates like this aren't pulled in automatically as a measure of safety and stability. I just updated pretalx.com to include this change.

tyrasd:
yeah, I meant on pretalx.com. Thanks! :)

PR #809, opened 2019-11-11
Cfp editor

rixx:
This PR adds the first step of a CfP editor: Being able to change the title and main text of every CfP step.


rixx:
Relates to #648

PR #812, opened 2019-11-18
✨ Allow plugins to add CfP steps

rixx:


orlando:
Nice this is great, I was doing this manually and had to modify the CFP steps, but now I can move that functionality to the plugin itself.

Thanks @rixx

rixx:
That's great to hear – if you want to share your code somewhere, that would be very cool, too!

Please note that I marked this API as unstable. I have no direct plans to change it, but I want to use it in a couple of plugins (which I'm currently working on) to see if the API makes sense. I jsut didn't want to promise stability before testing it first. That also means that if this API isn't sufficient for your purposes, there's still time to get it changed!

orlando:
@rixx I have a fork here https://github.com/orlando/pretalx/tree/od/iff with some changes we need. From the beginning I wanted everything to be in the plugin but I couldn't find a way to make it work. Part of the functionality is here in this plugin https://github.com/orlando/pretalx-iff-dif.

We are going to be adding more customizations, like tagging for speakers and submissions later on, but we have a really tight deadline for all this so I'm not sure what we end implementing at the end

PR #847, opened 2020-01-05
fix developer documentation link in README

rahmonov:
I was sad for a moment thinking the developer documentation is gone. Turns out it was just a change in the link :)


rixx:
Oh, oops! Thank you :)

Issue #523, opened 2018-11-07
It should be able to limit questions to one type of talks only


Nakaner:
## Problem you are facing

I would like to ask all speakers of one of our four types of talks at FOSSGIS 2019 to release their content under CC-BY. But I don't want to ask those how submit a workshop because their material will not be published.

## Possible Solution

The admin of a conference should be able to select which types of talks a questions should apply to.

rixx:
"type of talk" is a pretty wide area. We could restrict by tracks (which not all events use) or submission types, or both.

Nakaner:
The limitation to one or many submission types is implemented in the linked pull request which you are free to modify.

rixx:
Since tracks work differently, a couple of additional considerations would have to take place. That's why I opened up the question of which restriction should be implemented. Your PR is of course noted and will be merged in time, as long as we don't determine that questions should be restricted by track only.

rixx:
I think there is sufficient interest to warrant re-opening the issue.

Nakaner:
> The limitation to one or many submission types is implemented in the linked pull request which you are free to modify.

I do not know why I did not link the pull request back then but that's the link: https://github.com/pretalx/pretalx/pull/582 (I don't know how easy it can be merged nowadays because I have not followed Pretalx development for the last few months in detail)

rixx:
Yeah, I'll look into it – I changed a bunch of core CfP code recently which will probably require a couple of changes. Thank you!

rixx:
This is an issue that is well-suited for external contributors – let me know if I can be of any help, either via this issue or via support@pretalx.com.

## What this is about

pretalx provides a [Question](https://docs.pretalx.org/en/latest/developer/interfaces/models.html#questions-and-answers) model to get arbitrary information from speakers or reviewers.

pretalx also divides submissions by type, and (optionally) track. It would be useful to allow questions to be limited by certain submission types, just like they can already be limited to certain tracks.

## How to get started

First of all, leave a comment on this issue to make sure nobody else starts working on this issue at the same time!

Next, you need pretalx running locally on your machine. First, [fork](https://github.com/pretalx/pretalx/fork) the repository to your own GitHub account. Then clone it locally (or [set it as your remote](https://help.github.com/articles/configuring-a-remote-for-a-fork/) if you already cloned it earlier)

Follow the [developer setup guide](https://docs.pretalx.org/en/latest/developer/setup.html), and then play around with pretalx a bit.
[...]

PR #893, opened 2020-03-04
✨ Add ability to restrict questions by submission type

rahmonov:
This PR closes/references issue #523. It does so by …

## Checklist


rahmonov:
Hey @rixx. This is a WIP but it gets the job done. Could you please take a quick look and let me know if I am in the right direction? If so, I will next look at the unit tests for this.

src/pretalx/orga/templates/orga/cfp/question_form.html
```diff
@@ -64,6 +64,11 @@ <h2>
             {% bootstrap_field form.tracks layout='event' %}
         </span>
         {% endif %}
+        {% if form.submission_types %}
+        <span id="limit-submission">
```
rixx: Let's change this to a `class` here and five lines above, because `id`s are technically meant to be unique.

rahmonov:
@rixx made some changes. What do you think?

src/pretalx/cfp/flow.py
```diff
@@ -369,13 +369,21 @@ def is_applicable(self, request):
...
+            )
         ).exists()
 
     def get_form_kwargs(self):
         result = super().get_form_kwargs()
+        info_data = self.cfp_session.get("data", {}).get("info", {})
         result["target"] = ""
-        result["track"] = self.cfp_session.get("data", {}).get("track")
+        result["track"] = info_data.get("track")
+        result["submission_type"] = info_data.get("submission_type")
```
rixx: Hm, is this correct? Previously we used ``self.cfp_session["data"]["track"]``, and now we're using ``self.cfp_session["data"]["info"]["track"]``.

src/pretalx/static/orga/js/main.js
```diff
@@ -136,9 +136,9 @@ function question_page_toggle_view() {
 }
 
 function question_page_toggle_target_view() {
-  if ($("#limit-submission").length) {
+  if ($(".limit-submission").length) {
     setVisibility(
-      "#limit-submission",
+      ".limit-submission",
```
rixx: This won't work, I think, unless you modify the ``setVisibility`` function to iterate over the results of ``document.querySelectorAll``. ``document.querySelector`` only returns the first match.

src/pretalx/cfp/flow.py
rahmonov: I tested it multiple times with different types of questions. As long as I am testing correctly, it seems to work 😅.

I believe there was a bug here. I think that there reason it seemed to work is because `is_applicable()` would hide the questions page entirely so nobody saw the bug. But, it is my assumption and I am not 100% sure.

Can you please check this on your end?

src/pretalx/static/orga/js/main.js
rahmonov: Modified the `setVisibility`. I hope we are not constrained to a vanilla JavaScript and `$` is an option. Let me know what you think.

PR #898, opened 2020-03-14
docs: show runserver command for custom IP:port

ThomasWaldmann:
if you want to access pretalx development server not on localhost, but on another (LAN) IP, you need this.

due to tox being in between, it needs a special commandline that is not very intuitive, so we better have that in the docs than requiring everybody to find this out.

### DIFF
--- a/doc/developer/setup.rst
+++ b/doc/developer/setup.rst
@@ -96,6 +96,10 @@ To run the local development server, execute::
 Now point your browser to http://localhost:8000/orga/ – You should be able to log in and play
 around!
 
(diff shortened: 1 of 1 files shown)


rixx:
This feels like it's a good start to bloat the docs. It might be better to show in a generic way how to modify commands, and refer to the pretalx manage command documentation for flags etc.

(And this is definitely one of the downsides of using tox, and why I don't use tox myself.)

ThomasWaldmann:
The closing commit is not correct.

Issue #772, opened 2019-09-11
Allow organisers to leave a public comment with a release


rixx:
This comment would then be shown next to the list of changes, probably before.

rixx:
This is an issue that is well-suited for external contributors – let me know if I can be of any help, either via this issue or via support@pretalx.com.

## What this is about

pretalx allows organisers to schedule events, and then release public versions of these schedules. It would be great to be able to leave a note on these schedule changes.

## How to get started

First of all, leave a comment on this issue to make sure nobody else starts working on this issue at the same time!

Next, you need pretalx running locally on your machine. First, [fork](https://github.com/pretalx/pretalx/fork) the repository to your own GitHub account. Then clone it locally (or [set it as your remote](https://help.github.com/articles/configuring-a-remote-for-a-fork/) if you already cloned it earlier)

Follow the [developer setup guide](https://docs.pretalx.org/en/latest/developer/setup.html), and then play around with pretalx a bit.
Have a look at the places impacting this issue. Look at the code responsible for the issues (searching for relevant words or files should help you find your way around). We have an explanation of our [code structure](https://docs.pretalx.org/en/latest/developer/concepts.html) and the actual [data models](https://docs.pretalx.org/en/latest/developer/interfaces/models.html).


## How to do this

The Schedule model is in `pretalx/schedule/models/schedule.py`. Add a new nullable field called `comment`, then create a migration with the `makemigrations` command, and apply it with the `migrate` command.

Next, you'll want to include this field in the organiser area when releasing a schedule. Add it to the `fields` list of `pretalx.orga.forms.schedule.ScheduleReleaseForm`, and include it in `pretalx/orga/templates/orga/schedule/release.html`.

Finally, make sure this text is shown in the public changelog (`pretalx/agenda/templates/agenda/changelog.html`) and RSS feed at `pretalx/agenda/templates/agenda/feed/description.html`.

## Submit the PR

There are a number of tutorials on the Pull Request workflow online, for example this [short version](https://guides.github.com/introduction/flow/) by GitHub itself and this [more in-depth tutorial](https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github).

Remember to provide documentation and tests, if applicable. Run the tests before opening your PR, including style checks. We have a [contribution guide](https://docs.pretalx.org/en/latest/developer/contributing.html), which ought to prove helpful at this stage.

rahmonov:
Do you think it should be a `CharField` or a `TextField`?

[...]

PR #899, opened 2020-03-21
✨ Allow organizers to comment on a release

rahmonov:
This PR closes/references issue #772. It does so by adding a comment field to the schedule model.

image

## Checklist


rahmonov:
@rixx could you please take a look? If this looks good, I will fix the unit tests. Quite a bit of them are broken because of the changes to release_schedule and freeze methods.

src/pretalx/agenda/templates/agenda/changelog_block.html
```diff
@@ -1,4 +1,7 @@
 {% load i18n %}
+{% if schedule.comment %}
+  <p>{{ schedule.comment }}</p>
```
rixx: I'm thinking we should allow Markdown in the comment field, and then render it with our ``|rich_text`` template tag here.

src/pretalx/event/models/event.py
```diff
@@ -781,16 +781,17 @@ def get_date_range_display(self) -> str:
         """
         return daterange(self.date_from, self.date_to)
 
-    def release_schedule(self, name: str, user=None, notify_speakers: bool = False):
+    def release_schedule(self, name: str, comment: str, user=None, notify_speakers: bool = False):
```
rixx: Please make the comment optional, defaulting to ``None``, and move it to the end of the method signature – that way less tests should break, too.

src/pretalx/schedule/models/schedule.py
```diff
@@ -48,11 +56,12 @@ class urls(EventUrls):
         public = "{self.event.urls.schedule}v/{self.url_version}/"
 
     @transaction.atomic
-    def freeze(self, name: str, user=None, notify_speakers: bool = True):
+    def freeze(self, name: str, comment: str, user=None, notify_speakers: bool = True):
```
rixx: Please default to ``None`` here, as well, and put the comment to the end of the list of arguments. (I'm pretty sure we never call this method without explicit arguments in pretalx, but plugins may differ.)

src/pretalx/schedule/models/schedule.py
```diff
@@ -37,6 +38,13 @@ class Schedule(LogMixin, models.Model):
         max_length=190, null=True, blank=True, verbose_name=_("version")
     )
     published = models.DateTimeField(null=True, blank=True)
+    comment = I18nTextField(
+        help_text=_(
+            "This text will be shown in the public changelog and the RSS feed."
```
rixx: Could you please add ``phrases.use_markdown`` here? It's used in ``pretalx/submission/models/question.py`` in a similar place, for example.

src/pretalx/orga/views/schedule.py
```diff
@@ -141,6 +142,7 @@ def form_invalid(self, form):
     def form_valid(self, form):
         self.request.event.release_schedule(
             form.cleaned_data["version"],
+            form.cleaned_data["comment"],
```
rixx: Just a reminder, as per my other comments I'd prefer keyword arguments for this method, so please adjust the call here, too.

src/pretalx/orga/forms/schedule.py
```diff
@@ -1,10 +1,11 @@
 from django.forms import BooleanField, ModelForm, ValidationError
 from django.utils.translation import gettext_lazy as _
+from i18nfield.forms import I18nModelForm
 
 from pretalx.schedule.models import Schedule
 
 
-class ScheduleReleaseForm(ModelForm):
+class ScheduleReleaseForm(ModelForm, I18nModelForm):
```
rixx: ``I18nModelForm`` subclasses ``ModelForm``, so we only need ``I18nModelForm`` here.

PR #905, opened 2020-03-28
fix typo in referrer content

ThomasWaldmann:
https://lists.w3.org/Archives/Public/public-webappsec/2015May/0064.html

error from chromium console log:

Failed to set referrer policy:
The value 'strict-origin-when-crossorigin' is not one of 'always', ..., 'strict-origin-when-cross-origin', or 'unsafe-url'.
The referrer policy has been left unchanged.


rixx:
Thank you for catching this! Seems to be present in the last release, too, so this is pushing my plans for a new minor version release up by a bit.

I'm closing this PR, because as of the last major Django version, Django supports setting the referrer policy by way of the ``SecurityMiddleware`` (HTTP header instead of meta tags). I had applied the appropriate setting after upgrading to current Django, but forgotten to remove the meta tags – and faulty meta tags at that! I think HTTP headers are the preferable way of setting the policy, though.

Nevertheless, thank you for your contribution in catching + reporting + fixing the issue!

PR #962, opened 2020-07-07
Fix/ical error

rixx:


rixx:
I superseded this with some minor changes to speed things up, but thank you for the code regardless!

PR #984, opened 2020-08-01
Improvements to public talk/speaker page

glasnt:
A Pull Request in two parts: 

### Talk page: Add speaker bio

Much consideration was taken to this page, but given the existing translations and layout of the click-through speaker page, the minimal default profile information, just the bio has been added to the talk page.

Bio before Bio after

### Improve resolution of retrieved gravatar

Previously: the retrieved gravatar was the default 80x80 size, in a `.speaker-avatar` div of width 180

Now: retrieved avatar is 200x200, which is larger than the display size, but is still a useful size for clearer display.
Avatar before
Avatar after 

## How Has This Been Tested?

Testing in local instance

See above

## Checklist

### DIFF
--- a/src/pretalx/agenda/templates/agenda/speaker.html
+++ b/src/pretalx/agenda/templates/agenda/speaker.html
@@ -30,7 +30,7 @@
             {% if profile.user.get_gravatar or profile.user.avatar %}
             <img width="100%"
                  {% if profile.user.get_gravatar %}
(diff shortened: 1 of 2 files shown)


glasnt:
I tested this with a 1000 character bio with half a dozen linebreaks and it still looked okay. 
I wouldn't want to add a limit in the config just for a display, and I wouldn't want to truncate bios on display.

rixx:
Alright, thank you for checking!

Issue #994, opened 2020-08-25
SMTP Password field visibility


momorientes:
## Current Behavior
When reviewing E-Mail settings the Password field is currently displayed empty even if a valid password is stored.

## Expected Behavior

If no password is stored the field should be empty, if a password is stored it should be indicated that a value has been set (i.e. by displaying a sequence of asterisks)

## Steps to Reproduce

1. Go to E-Mail Settings for an event
2. enter a smtp password & save

## Context

We often copy from previous events and it's never clear if the password has been correctly copied or not (or if it is our mailserver acting up).

chriswolfdesign:
I'm very new at open source so I'm a bit nervous about offering, but I could try to handle this one if you would like?

rixx:
Sure!

To set up pretalx, you can follow the [developmtent setup guide](https://docs.pretalx.org/developer/setup.html). Then, you will want to look at the email settings form – the form definition can be found in `pretalx/orga/forms/event.py` as `MailSettingsForm`. I think the best idea would be to override the `placeholder` of the `smtp_password` field to something like `************` if an SMTP password is already set, and leave it unchanged otherwise. Let me know if you need any help on the way!

(The reason that this is necessary is that this settings form can be accessed by every organiser, but we don't want to leak the email password between them, so we don't actually expose the password here.)

chriswolfdesign:
Great, I work a full time software development job so I may not be able to get around to it until this weekend but I'd really like to start contributing to projects.  Quick question about the expectation.  Do we want the number of asterisks to fit the length of the password?  So if the password was "foo" would we want the password to display as ***?  Or do we want the length of the asterisks to be the same, regardless of the length of the password?

rixx:
We should keep the number of asterisks constant, to avoid leaking any information about the password.

chriswolfdesign:
Sorry this took longer than I had expected to get back to you guys.  I had some unexpected guests this weekend.  I have finished the functionality but I have a few concerns.  First, whenever I run existing tests these two fail:

* test_agenda_schedule_export: test_html_export_language
* test_cfp_user: test_can_change_locale

I'm not sure how to go about making those two tests pass and I'm also not sure how to go about unit testing my own code I wrote.  I'm still quite new at Django.  Could anybody provide a bit of guidance on these topics?  Sorry for any inconvenience.

rixx:
[...]

PR #1000, opened 2020-09-01
🎀 Add placeholder to email settings password field

chriswolfdesign:
When inheriting an event, it was unclear if this event already had a password set up.  Now if there is no password, the password's placeholder will be "Password". If there is a password, the placeholder will be "********" (8 asterisks) these eight asterisks are stored as a global constant for convenient changing

This PR closes/references issue #994. It does so by completing functionality discussed above

## How Has This Been Tested?

User testing.  I created new events and changed the passwords.  I also inherited events that both did and did not have existing passwords.

Mac OSX and venv

I did not notice any other changes in flexibility

## Checklist

- [X] My code follows the code style of this project. (I believe so)
- [X] My change requires a change to the documentation. (This would be up to maintainer's discretion)
- [X] All new and existing tests passed. (Only failing tests seem to be environment driven rather than code driven)

### DIFF
--- a/src/pretalx/orga/forms/event.py
+++ b/src/pretalx/orga/forms/event.py
@@ -19,6 +19,7 @@ from pretalx.common.phrases import phrases
 from pretalx.event.models.event import Event, Event_SettingsStore
 from pretalx.orga.forms.widgets import HeaderSelect, MultipleLanguagesWidget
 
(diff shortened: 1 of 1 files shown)


rixx (approved):
Nice work, thank you!

PR #1027, opened 2020-11-05
Fix display of date on talk page

adamzammit:
When the date stored in the database for a talk (eg in UTC time) is the day before (eg 1 December 2020), but the talk will actually take place on 2 December 2020 due to the timezone difference, the talk page will display the date of 1 December 2020 instead of 2 December 2020.

This change makes sure the date object doesn't lose it's timezone component by being converted to a simple date object and instead uses a "Y-m-d" formatter to display it.

This fix is only for the talk template: I suggest looking over the template code for places where this is used:

.date|date:"c"

and replace it with:

|date:"Y-m-d"

## How Has This Been Tested?

Code updated on local copy of site and results appear to match expected date display

## Checklist

- [X ] My code follows the code style of this project.

### DIFF
--- a/src/pretalx/agenda/templates/agenda/talk.html
+++ b/src/pretalx/agenda/templates/agenda/talk.html
@@ -42,7 +42,7 @@
         </div>
         <small>
             {% if talk_slots|length == 1 %}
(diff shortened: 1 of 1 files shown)


rixx (approved):
Thank you very much, both for the contribution and the detailed comment

PR #1034, opened 2020-11-26
Fix Link-Example

MaZderMind:
Our users were adding dots to the End of their URLs, according to the examples, and they did not work.


rixx:
Oops, thank you for catching this, and sorry for having to fix those links!

rixx (approved):

Issue #1082, opened 2021-03-24
Event logo is not scaled to fit header


Kunsi:
When uploading an Event logo, the logo is embedded as-is, without any scaling. This results in the page content being pushed down very far when uploading a large image.

Alternatively pretalx could provide a hint about what image size would work best.

rixx:
Event images are now assured to be smaller in height than the background header, fixed at a maximum of 150px.

Issue #1084, opened 2021-03-29
Show avatar image upload size limit in help text


rixx:
This leads to confusion for submitters, so we should have a size hint for image fields.

AthinaDavari:
@rixx , me and my colleague @NataliaKat want to contribute to this issue.  

The changes we want to make are:
1.	In src/pretalx/static/cfp/scss/_layout.scss file, we will change #event-logo. To be more specific, we will add max height to this rule and delete the rule height:auto. We will also delete width: 100%.
2.	We will add a help text in two parts of the of the software: event’s settings (events' logo) and in new event’s form (step 4 : logo). The help text will we like this: “If the image is taller than x px, it will automatically be adjusted to x px.”.  (The suggested max height is 150 px.)

These changes will also satisfy issue #1082.

We are first time contributors and we want to contribute as part of Applied Software Engineering university course. We would like to make further contributions, so any help or suggestion is welcome.

rixx:
Sorry for not responding earlier, I was away – your efforts are very appreciated! Happy to have your contributions, and ping me whenever you have any questions or want any help.

PR #1096, opened 2021-04-11
Fix help texts and logo's scaling

AthinaDavari:
This PR closes issues #1082 & #1084. It does so by fixing help texts for logo and header images in event's settings. Also, we changed  #event-logo height rule to max-height and deleted width rule.

## Screenshots:

Screenshot (279)

## Checklist

- [X] All existing tests passed.

### DIFF
--- a/src/pretalx/orga/forms/event.py
+++ b/src/pretalx/orga/forms/event.py
@@ -37,19 +37,19 @@ class EventForm(ReadOnlyFlag, I18nModelForm):
     )
     logo = ImageField(
         required=False,
-        label=_("Header image"),
+        label="",
         help_text=_(
-            "If you provide a header image, it will be displayed instead of your event's color and/or header pattern "
-            "on top of all event pages. It will be center-aligned, so when the window shrinks, the center parts will "
-            "continue to be displayed, and not stretched."
+            "If you provide a logo image, we will by default not show your event's name and date in the page header. "
+            "We will show your logo in its full size if possible, scaled down to the full header width."
         ),
     )
     header_image = ImageField(
         required=False,
-        label=_("Header image"),
+        label="",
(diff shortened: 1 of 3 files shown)


rixx (commented):
Thank you for your PR! Could you show how this changes the look of existing events with larger images? We probably need to warn users about changing looks of their events.

AthinaDavari:
Thanks for responding.
The changes we made are for logos with height more than 200px. In these cases, we scale down the logos to 200px.
### Before changes:

Στιγμιότυπο οθόνης (275)

### After changes:

Στιγμιότυπο οθόνης (276)

rixx (approved):
I wonder if we shouldn't contain the logo to the header entirely, once we make this switch. A logo looking past the header just by a bit looks fairly awkward to me.

Doesn't need to be decided now though – I'll merge it as is. Thank you for your contribution!

src/pretalx/orga/forms/event.py
```diff
@@ -37,19 +37,19 @@ class EventForm(ReadOnlyFlag, I18nModelForm):
...
     logo = ImageField(
         required=False,
-        label=_("Header image"),
+        label="",
         help_text=_(
-            "If you provide a header image, it will be displayed instead of your event's color and/or header pattern "
-            "on top of all event pages. It will be center-aligned, so when the window shrinks, the center parts will "
-            "continue to be displayed, and not stretched."
+            "If you provide a logo image, we will by default not show your event's name and date in the page header. "
+            "We will show your logo in its full size if possible, scaled down to the full header width."
```
rixx, line 43: "scaled down to the full header width" isn't really accurate, as the 200px height will be the limiting factor for most events, I think. I'll fix this in a follow-up commit.

PR #1117, opened 2021-06-04
📚 Update section for email in development setup

cfra:
When running in development mode, the server is using Django's console
email backend and prints mails to stdout instead of sending them via
SMTP.

The mail configuration is also not loaded by `settings.py` leading to the
proposed port setting of 1025 to be ignored, and default settings
showing up in the admin information, which was a bit confusing to me.

Python's debugging SMTP server was still news to me, so I think it is
useful to keep that information around, in case somebody needs to test
interactions via SMTP.

### DIFF
--- a/doc/developer/setup.rst
+++ b/doc/developer/setup.rst
@@ -144,16 +144,17 @@ afterwards to format that file.
 Working with mails
 ^^^^^^^^^^^^^^^^^^
 
-If you want to test emails in your development setup, we recommend starting
-Python's debugging SMTP server in a separate shell and configuring pretalx to
-use it. The debugging SMTP server will print every email to its stdout.
+When running in development mode, Pretalx uses Django's console email backend.
+This means the development server will print any emails to its stdout, instead
+of sending them via SMTP.
 
-Add this to your ``src/pretalx.cfg``::
+If you want to test sending event emails via a custom SMTP server, we recommend
+starting Python's debugging SMTP server in a separate shell::
 
-    [mail]
-    port = 1025
+    python -m smtpd -n -c DebuggingServer localhost:1025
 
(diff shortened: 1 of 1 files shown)


rixx (changes requested):
Thank you! Could you change this to also keep the config snippet in there? I like to make stuff like that easy to copy and paste, which "by specifying ``localhost``" isn't really.

cfra:
I removed it intentionally, because it does not work, because these settings are ignored when debugging is enabled:

```python
# from settings.py

## EMAIL SETTINGS
MAIL_FROM = SERVER_EMAIL = DEFAULT_FROM_EMAIL = config.get("mail", "from")
if DEBUG:
    EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
else:
    EMAIL_HOST = config.get("mail", "host")
    EMAIL_PORT = config.get("mail", "port")
    EMAIL_HOST_USER = config.get("mail", "user")
    EMAIL_HOST_PASSWORD = config.get("mail", "password")
    EMAIL_USE_TLS = config.getboolean("mail", "tls")
    EMAIL_USE_SSL = config.getboolean("mail", "ssl")
```

The last section about "specifying localhost" relates to event settings that are configured using the form in the back end as far as I understood it, and not via the settings file.

As far as I can tell, the config snippet I remove will only work if `DEBUG = False` is configured, which I assumed is not commonly done in a development setup.

I think it would be good to point out that the included snippet will not work with `DEBUG = True`, if it is included in the dev setup documentation, because it was somewhat confusing to me.

What do you think?

rixx:
Apologies, I misunderstood parts of this PR. I jumped to thinking about making the behaviour described in the documentation work, rather than changing the documentation. Still, working documentation is better than incorrect documentation, so I'll merge this with thanks!

rixx (approved):

PR #1129, opened 2021-07-02
✨ export public answers within schedule.json 

kuhball:
Public answers currently are only exported for logged in users. Using the schedule.json within different systems it's nice to have the questions available for further usage. 

## How Has This Been Tested?

Tested using the development environment creating multiple questions / events. Also tested within a container setup. 

## Checklist

- [x] All new and existing tests passed. - the same amount of tests fails :wink: 

### DIFF
--- a/src/pretalx/schedule/exporters.py
+++ b/src/pretalx/schedule/exporters.py
@@ -233,7 +233,19 @@ class FrabJsonExporter(ScheduleData):
                                         for answer in talk.submission.answers.all()
                                     ]
                                     if getattr(self, "is_orga", False)
-                                    else [],
+                                    else [
+                                        {
+                                            "question": answer.question.id,
+                                            "answer": answer.answer,
+                                            "options": [
+                                                option.answer
+                                                for option in answer.options.all()
+                                            ],
+                                        }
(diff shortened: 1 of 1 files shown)


rixx:
The schedule.json is meant to integrate with tools compatible with frab, so adding features that frab doesn't support seems like a bad idea: Now there are two standards etc. I don't imagine the apps using this interface (giggity, schedule app, VOC) would be very interested in this extension, though I'm happy to talk it through if I'm wrong.

Answers can already be exported via the public API or the organiser-only JSON/CSV export. The schedule.json and schedule.xml exports exist only for frab compatibility, so all changes to them should be talked over with the frab team and/or the VOC team (who are the main maintainers and users of this compatibility features). This goes especially since the schedule.json should be a mirror of the schedule.xml as much as possible, and there exists a validation schema for the schedule.xml format.

kuhball:
I understand your arguments, didn't give this a big thought due to the answer field already available within the schedule.json. This should definilitly be implemented within the schedule.xml. 

Looking at the answer section it"s hard to use this section within apps due to the id for mapping the question to the answer. Should have given this more thought beforehand.

PR #1136, opened 2021-07-12
fix redis port number in docs

Findus23:
I noticed that the port number in the config description is confusing as 6379 is the "normal" port number for redis

I'm also a bit confused by these lines:
https://github.com/pretalx/pretalx/blob/ce106b7b140c0e0f2178bf7387acbedc4c2c809a/doc/administrator/configure.rst#L270-L276

This setting seems to be a boolean, which would mean that a "sensible value" of a string doesn't make any sense.

## Checklist


rixx (approved):
Absolutely a typo, thanks!

PR #1137, opened 2021-07-13
no quotation signs in config file

Findus23:
related to #1136

When I tried to set up Pretalx yesterday I got stuck for quite some time getting mysterious errors like `ValueError: Redis URL must specify one of the following schemes (redis://, rediss://, unix://)` or `ValueError: Incorrect timezone setting: "Europe/Berlin"`. 
After missing the forest for the trees for some time, I finally noticed that I was always entering `backend = 'redis://127.0.0.1/1'` and similar values in the config file, but python ini files don't need any quotes and in fact then use the quotes as part of the string which makes the error messages really confusing.

So I think this should be made a bit more clear in the docs in case someone else is coming across the same issue.

 

## Checklist


rixx (approved):
Thank you for mentioning the problem AND for providing the solution!

PR #1165, opened 2021-09-12
🎀 show sessions in full-width too

cambid:
If running an event with more than 3 days the session screen is fixed width and not full-width as the schedule.

An example of the problem is accessible under https://fahrplan.privacyweek.at/pw21/talk/#

## How Has This Been Tested?
local tests in multiple browsers. 

## Checklist

### DIFF
--- a/src/pretalx/agenda/templates/agenda/schedule.html
+++ b/src/pretalx/agenda/templates/agenda/schedule.html
@@ -5,7 +5,7 @@
 
 {% block body_tag %}{% if not request.event.settings.schedule_display or request.event.settings.schedule_display == "proportional" %} class="body-schedule"{% endif %}{% endblock %}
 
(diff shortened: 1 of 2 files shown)


rixx:
If I understand you correctly, you want to fix the fact that the header scrolls sideways when there are too many days?

I don't think setting the session list to always display at 100% width can be the solution – that would look horrible by default, since stretching the session blocks over full-hd (or larger!) width makes them hard to read and to parse.

cambid:
Yes the side-scrollbar in the day-list on larger screens is our problem. I see the problem in readability with the 100% width an I'm open for any suggestions how to fix the original problem.

rixx:
The schedule now tries to shorten the tab entries for events that run longer than five days and are displayed in the list view (or the grid view with less than three rooms).

PR #1209, opened 2021-11-08
Fix incorrect HTTP header name

lujoga:
`Referrer-Options` should be changed to `Referrer-Policy` (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy)


rixx (approved):
Thanks!

PR #1213, opened 2021-11-17
Add Italian locale

pretalx-translations:
Translations update from [Weblate](http://translate.pretalx.com/projects/pretalx/pretalx/) for pretalx/pretalx.

Current translation status:

Weblate translation status


rixx (commented):
Not adding Italian to list of available translations yet

PR #1230, opened 2021-12-14
Restructure settings access

rixx:
## Reasoning

Hierarkey is a hierarchical key-value store, but pretalx has long made a habit out of using it just as a generic key-value store, because writing migrations for every tiny setting is exhausting and now fun. This leads to an unreasonable increase in single DB lookups though, that we really should avoid.

As Django now supports JSON fields regardless of underlying database, we're just shoving nearly all settings into JSON fields on `Event` and `CfP` for now. Global settings remain in hierarkey, naturally, as do i18n strings because they a) are annoying to properly parse and b) don't currently concern pages with a strong performance impact. If this changes in the future, well, sucks to be me and I'll have to write this migration again.

## Hierarkey

Hierarkey is not going to be phased out of pretalx: It's way too useful for that, and plugins rely on it. We'll try to stop using it as a generic key-value store, though.


rixx:
Ignoring the coverage changes, as a lot of the misses is just us vendoring a hierarkey form for now.

PR #1312, opened 2022-05-17
fix setting a field to required in CfP not having any effect

codingcatgirl:
Setting avatars to be required had no effect in our instance.

Upon further inspection, i found that while in the settings, the value `"required"` is stored in this case, `is_field_required()` checked for `"require"` (without d) instead.

This PR fixes this typo.

Drawback: While nearly all CfP fields that are required by default use the value `"required"`, the field `"content_locale"` used `"require"` (which this PR also fixes). If this default value has persisted for some users, this field might no longer be required now (didn'tread into the code enough to check if this method is called for all of the CfP form generation).

Now, setting avatars to be required and not filling out that field actually throws a validation error as expected.

## Checklist

### DIFF
--- a/src/pretalx/submission/models/cfp.py
+++ b/src/pretalx/submission/models/cfp.py
@@ -50,7 +50,7 @@ def default_fields():
         "image": {"visibility": "optional"},
         "track": {"visibility": "do_not_ask"},
         "duration": {"visibility": "do_not_ask"},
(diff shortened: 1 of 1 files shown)


rixx (approved):
Oh wow. That broke as part of a recent rewrite into this settings structure, I believe.

Thank you for the fix!

Issue #1334, opened 2022-07-19
json export doesn't contain time_zone_name


aNullValue:
## Current Behavior

When viewing frab-compatible XML, attribute schedule->conference->time_zone_name exists and is set to a logical value. But when viewing frab-compatible JSON, no such attribute exists. 

(Same problem with "generator" name and version not being in the json, and speaker biography not being in XML, but one problem per request I suppose. )

## Expected Behavior

Having identical information in both file formats.

## Steps to Reproduce

1. View https://scheduler.hope.net/new-hope/schedule/export/schedule.xml
2. View https://scheduler.hope.net/new-hope/schedule/export/schedule.json
3. Observe missing "time_zone_name" value anywhere in the json file.

## Context

I'm trying to process only the JSON file, to get as much info as possible. I intend to use it as a sync-source connected to the conference management software that powers HackerTracker (a conference app for smartphones). 

## Your Environment

HOPE appears to be running pretalx 2.3.1, the latest release. I'm just attempting to consume the XML and JSON files. I am not affiliated with HOPE.

saerdnaer:
Actually frab includes the time zone name in the JSON version...

https://fahrplan.events.ccc.de/rc3/2020/Fahrplan/schedule.json / https://frab.cccv.de/en/rc3/public/schedule.json
```
{
"schedule": {
  "version": "1.7",
  "base_url": "https://fahrplan.events.ccc.de/rc3/2020/Fahrplan/",
  "conference": {
    "acronym": "rc3",
    "title": "Remote Chaos Experience",
    "start": "2020-12-27",
    "end": "2020-12-30",
    "daysCount": 4,
    "timeslot_duration": "00:10",
    "time_zone_name": "Europe/Berlin",
    …
```

I also started to create a JSON schema at https://github.com/voc/schedule/tree/master/validator/json , but of course `time_zone_name` is still missing there... because I was thinking about additional attributes like color, keywords, etc. – c.f. https://github.com/voc/streaming-website/blob/master/docs/config-schema.json#L76-L102

I think we (aka CCC / VOC) should decouple the format binding to frab, and simply provide some a python package (e.g. `voctosched`) containing the XSD, JSON-Schema and maybe even Python data models…

rixx:
I'm happy to integrate whatever the community decides on. We might move the frab exporters to a plugin at that point, if there's significant overhead – it's easier for the community to maintain then, and events can decide if they want/need the often confusing "frab exporters" in their export list. 

[...]

PR #1338, opened 2022-07-25
Validate pretalx frab json exports with VOC schema

rixx:
Continuing the conversation at #1334: VOC has a JSON schema that we can use to comply against a shared set of expectations, instead of being surprised by frab changes.

Currently waiting for changes in the JSON schema.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,7 @@ dev = [
   "flake8-bugbear",
   "freezegun",
   "isort",
+  "jsonschema",
   "lxml",
   "pytest",
   "pytest-cov",
--- a/src/tests/agenda/test_agenda_schedule_export.py
+++ b/src/tests/agenda/test_agenda_schedule_export.py
@@ -9,6 +9,7 @@ from django.core.management.base import CommandError
 from django.test import override_settings
 from django.urls import reverse
 from django_scopes import scope
+from jsonschema import validate
 from lxml import etree
 
 from pretalx.agenda.tasks import export_schedule_html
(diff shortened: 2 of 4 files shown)


saerdnaer:
Sorry for the delay, the JSON schema now allows `UTC` as valid time zone name (and regex had another bug, too) – Thanks for pointing me to that issue!

rixx:
No worries, it's not like it's urgent – thanks for pushing this forward!

It seems the remaining failing check is for `room.guid`, which the schema expects to be present, but which is currently optional in pretalx, iirc for backwards compatibility reasons. No strong preference what to do here (though adding required room guids to pretalx requires writing three db migrations and I'm lazy …).

rixx:
It seems that the VOC schema currently requires room GUIDs, while these are an optional feature in pretalx.

At this point, I'm once again thinking about moving all the frab/VOC exports to a separate plugin, and that plugin could also be responsible for generating room GUIDs where necessary. Maintaining this in pretalx-core is looking increasingly like a maintenance burden due to lack of standardisation etc.

saerdnaer:
I would provide an seperate PR which generates the room guid from the pretalx domain name and the internal room id in the database – this should resolve the issues we had with the integration tests with the previous random based room guids.

rixx:
Sure! I'll consider this PR dormant until then.

saerdnaer:
@rixx When you update the JSON Schema from the source, this tests should become green now.

rixx:
You'll probably want to rebase this PR (and fix the merge conflict) to see if tests are passing now.

Kunsi:
This branch is in the official `pretalx` organization, i cannot rebase this because i lack committer access.

rixx:
> This branch is in the official `pretalx` organization, i cannot rebase this because i lack committer access.

Ah yeah, I somehow grouped this mentally under all the other VOC PRs (though in any case, I was addressing saerdnaer incorrectly, didn't mean to make you do anything!)

rixx:
> When you update the JSON Schema from the origin, this tests should become green now.

Nope. pretalx is returning a null-value for room GUIDs, whereas your change made the GUID field optional – so jsonschema correctly fails, because `None` is not a string value, which is intended for `room.guid`. Still blocked on either merging the room.guid feature branch or pulling all frab compatibility into a separate plugin, depending on what happens first.

PR #1360, opened 2022-10-13
switch from git to https

yvovandoorn:
This switches the command for the `commit` version to use `https` instead of `git`. Using the unsecured `git://` protocol has been disabled in 2021 (per https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git).


rixx (approved):
Oops, thanks!

PR #1362, opened 2022-10-17
Add availability support to CfP

fkusei:
This feature allows organizers to limit the availability time selection for the speaker when submitting their talk.

> **Disclaimer:**
> This feature has been done as sponsored development for our company.
> We are currently determining if the original author wants to be disclosed.

## How Has This Been Tested?
This has been running fine on our pretalx instance for the last couple of months, however that instance was not updated since april. We are currently in the process of testing wether the CfP workflow works when using this enhancement rebased on current master.

(will follow)

## Checklist


src/pretalx/schedule/forms.py
```diff
@@ -145,10 +206,37 @@ def clean_availabilities(self):
...
+        # sort availabilities, so we can merge ones that touch or overlap
+        rawavailabilities.sort(key=itemgetter("start"))
+
+        tz = pytz.timezone(self.event.timezone)
+
+        availabilities = []
+        for rawavail in rawavailabilities:
+            if self.availabilities_within:
+                # check if this availability entry is within availabilies_within
+                for availability in self.availabilities_within:
```
rixx, line 223: The `Availability` class already provides methods to find overlaps and inclusions, would probably be easier (less code or duplication) to use those.

fkusei:
Thanks for the quick reply. I will check whether we will be able to make those changes, or if we have to roll back the changes in our system to be able to use pretalx upstream in the future. I'll be back with future information.

fkusei:
Having thought of this for some time, i think it would be best to add a 'time' value to the 'conference start' and 'conference end' fields in the event configuration.

If we tied speaker availability to room availability, we'd have to have room information available prior to creating the CfP. I don't think this would be the best solution, because sometimes the room availability is simply not known at that time.

So i think the time values for conference start/end would be the best solution. However, i'm fairly sure we can't make these changes ourself in a timely manner. Should i create a issue somewhere?

PR #1445, opened 2023-03-13
fix crontab install step documentation

kvbik:
`pip install --user` is used through the install documentation, but there is no `python` binary in the user level isolation. In contrast from `gunicorn` or `celery` scripts.

I've run into this, when someone else without a deeper knowledge of python ecosystem installed Pretix on a server. They were just blindly following administration documentation but `/var/pretalx/.local/bin/python` clearly does not exist.

Not sure if my documentation update is enough, but at least I wanted to report this! Thanks for all the great work on Pretalx and congrats for the release :)

## Checklist


rixx (approved):
Thank you! That's absolutely a leftover from past installation methods.

PR #1453, opened 2023-03-25
fix: handle team without members in tests

### LINKED ISSUE #1452: Tests are failing both locally and in Github Actions
### Problem and impact

https://github.com/pretalx/pretalx/actions/runs/4389337240 - broken by https://github.com/pretalx/pretalx/commit/52b674b93960d73a2c05eea80058726009508fbd

### Expected behaviour

_No response_

### Steps to reproduce

1.
2.
3.
4.

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

_No response_

MikkCZ:
fix #1452 

This PR closes/references issue #1452. It does so by handling teams without `members` field. I am not 100% sure though if it's a valid case in the real operations and it's not test fixtures that are lacking.

## How Has This Been Tested?
Run `PRETALX_CONFIG_FILE=tests/ci_sqlite.cfg py.test -nauto -p no:sugar --cov=./ --cov-report=xml --reruns 3 tests --maxfail=100` locally and the three tests failing in Github Actions passed.

## Checklist

### DIFF
--- a/src/pretalx/orga/views/organiser.py
+++ b/src/pretalx/orga/views/organiser.py
@@ -65,7 +65,10 @@ class TeamDetail(PermissionRequired, TeamMixin, CreateOrUpdateView):
     @context
     @cached_property
     def members(self):
-        return self.team.members.all().order_by("name")
+        if hasattr(self.team, "members"):
+            return self.team.members.all().order_by("name")
+        else:
+            return []
 
     def post(self, *args, **kwargs):
         if self.invite_form.is_bound:


rixx (commented):
I agree that this is probably the most pythonic solution, but I don't like it: It's completely unhelpful to the reader, who is invited to figure out when a team has no `members`.

(Of course, it's when there is *no team*, but that's not particularly obvious unless if you're familiar with the code. Closing,will replace with `if not self.team`. Yes, this won't catch cases when there is a `team` which doesn't have members. Yes, I'd consider that a *good* thing, as that should never happen and we need to know if it does.)

Issue #1387, opened 2022-11-26
Prevent all scores from being marked as independent


rixx:
Pretalx doesn't deal well when all score categories are independent – and it doesn't make that much sense: independent from what? The total score is then shown as always zero, which is confusing and pointless.

rixx:
Go for it!

Pradhvan:
Apologies for the later reponse on the isssue. 

I cloned the project and played around with it locally. 

Got an understanding of  what an independent review is, how an org memember can created a ReviewScoreCategory that is independent and  how it shows up as their own column in the review dashboard.

The approach I am confused is with having a query like  

```Python
# Check if any previously saved score category has at least one category that has is_independent=False

ReviewScoreCategory.objects.exclude(pk=self.pk).filter(is_independent=False, event=self.event).exists()
```

1. Either in the `save()` of `ReviewScoreCategory` model. Should I raise an exception when it does not find an independt review ?

2. Or have it as a method so that it can be called in the `views.py::EventReviewSettings()::def save_scores()`` method where we save score.

Since deleting a score category is handeled in seprate API call which calls `ScoreCategoryDelete` we would have to update that class too. 

#### Things I tried locally

* Added the query mentioned above in the both model's save() and delete method. 
* Tested the approach in createting paris of independent score category and normal score categorys. Deleting one normal score category, keeping only independent. Trying to delete the idependent one in the pair. 

Though looks like I might be missing something in my logic.  

Just want some pointers on how to move forward, @rixx if you get some time can you briefly look at this and point to the right direction.Thank you. 😄

rixx:
That's looking pretty good to me! I think the puzzle piece you're missing is that the logic is probably best placed in `ScoreCategory.delete()`. I think raising a `ValidationError` would be best here, which we can then expose in forms, views and APIs equally.

(hope that's helpful! it really looks like you're mostly done already :smile:  )

PR #1467, opened 2023-04-19
Bug:Prevent all scores to be marked as independent

Pradhvan:
* Add check `ReviewScoreCategory::save()` that prevents saving all indepent score category.

* Overide `ReviewScoreCategory::delete()` and add a check to only delete ScoreCategory when a non independent score category exists.

* Raises `ValidationError` if both `save()` and `delete()` check fails.

Closes https://github.com/pretalx/pretalx/issues/1387

## Checklist

### DIFF
--- a/src/pretalx/submission/models/review.py
+++ b/src/pretalx/submission/models/review.py
@@ -41,13 +41,17 @@ class ReviewScoreCategory(models.Model):
         for review in event.reviews.all():
             review.save(update_score=True)
 
+    @property
+    def _validate_independence(self):
+        return (
+            ReviewScoreCategory.objects.exclude(pk=self.pk)
+            .filter(is_independent=False, event=self.event)
+            .exists()
+        )
+
     def save(self, *args, **kwargs):
         if self.is_independent:
-            if (
-                not ReviewScoreCategory.objects.exclude(pk=self.pk)
-                .filter(is_independent=False, event=self.event)
-                .exists()
-            ):
+            if not self._validate_independence:
                 raise ValidationError(
                     _("Atleast one non-independent score category is required."),
                 )
@@ -55,12 +59,7 @@ class ReviewScoreCategory(models.Model):
         return super().save(*args, **kwargs)
 
     def delete(self):
-        if (
-            self.is_independent
-            and not ReviewScoreCategory.objects.exclude(pk=self.pk)
-            .filter(is_independent=False, event=self.event)
-            .exists()
-        ):
+        if not self._validate_independence:
             raise ValidationError(
                 _("Atleast one non-independent score category is required."),
             )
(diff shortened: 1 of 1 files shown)


Pradhvan:
@rixx whenever you get time can you please check this approach. 

I added checks in both `save()` and `delete()`. 

Have tests the scenario:

1. Tried to save two independent score category. It raised an error.
2. Delete Independent score when no non Independent score is present. 
3. Save one non independent score, mark it as independent and save it. Gives validation error. 

😄  If the approach seems good, I can start adding unit tests in `test_review_model.py`.

src/pretalx/submission/models/review.py
```diff
@@ -42,9 +43,29 @@ def recalculate_scores(cls, event):
...
+            ):
+                raise ValidationError(
+                    _("Atleast one non-independent score category is required."),
+                )
             self.weight = 0
         return super().save(*args, **kwargs)
 
+    def delete(self):
+        if (
+            self.is_independent
```
Pradhvan, line 59: I think `self.is_independent` because if you have two reviews

1. One Non-Independent - score
2. One Independent - grammer

If you delete `score` this would let you delete that review score since it's not an independent review hence does not got inside the check.

rixx (commented):
Thanks!

Looking good at first glance, just one note: It would be nice to remove the code duplication, by moving the shared code into a `_validate_independence` method.

Pradhvan:
> Looking good at first glance, just one note: It would be nice to remove the code duplication, by moving the shared code into a `_validate_independence` method.

Added a `def _validate_independence(self)` method, marked it as property since if I used it without a property the function was returning a `bound method` error. 

Moving on to unit tests, will add tests in a new file `test_review_score_category_model.py`.

src/pretalx/submission/models/review.py
```diff
@@ -40,11 +41,30 @@ def recalculate_scores(cls, event):
         for review in event.reviews.all():
             review.save(update_score=True)
 
+    @property
```
rixx, line 44: What's the reason for this method to be a property? It's something that's called, and a verb, so semantically and logically it should be a method, imo.

rixx (commented):
Looking good apart from tests and one small quibble!

rixx:
Still interested in working on this, @Pradhvan ?

src/pretalx/submission/models/review.py
Pradhvan: Oh! I get that the reason I added it as a property as 

> Added a def _validate_independence(self) method, marked it as property since if I used it without a property the function was returning a bound method error.

But will work on it how this can be made into a normal method. 😄

Pradhvan:
> Still interested in working on this, @Pradhvan ?

Hey @rixx sorry for the late reply. Had been busy with job interviews so could not get much time on this. 😅 

I have a couple of days off till 8th let me see if I can fix this before that.

If it's urgent fix I get that, I would be happy if someone else can work on the PR based on this PR. 
 
 I hope that works ?

PR #1471, opened 2023-04-27
Fix Czech language flag

oskar456:
Czech language has ISO 639-1 code of `cs` while the proper flag is associated with the Czech Republic, that is ISO 3166-2 code `CZ`. This is a simple fix to display proper flag for Czech language.


rixx (approved):
Thanks!

PR #1486, opened 2023-05-14
Greek translation fixes

Sugarv:
A pull request which fixes some Greek translation errors, as well as improving some terms.


rixx (changes requested):
There were simultaneous changes to the Greek translation via [translate.pretalx.com](https://translate.pretalx.com/projects/pretalx/pretalx/), which took precedence. Could you please sort out the merge problems? Thanks!

Issue #399, opened 2018-03-31
Cache or proxy gravatar images


rixx:
This is an issue that is well-suited for external contributors – let me know if I can be of any help, either via this issue or via support@pretalx.com.

## What this is about

pretalx allows users to opt in to fetch their profile picture from gravatar. Due to privacy concerns, it would be better to either save (and refresh!) gravatar images locally or to provide a proxy.

## How to do started

First of all, leave a comment on this issue to make sure nobody else starts working on this issue at the same time!

Next, you need pretalx running locally on your machine. First, [fork](https://github.com/pretalx/pretalx/fork) the repository to your own GitHub account. Then clone it locally (or [set it as your remote](https://help.github.com/articles/configuring-a-remote-for-a-fork/) if you already cloned it earlier)

Follow the [developer setup guide](https://docs.pretalx.org/en/latest/developer/setup.html), and then play around with pretalx a bit.
Have a look at the places impacting this issue. Look at the code responsible for the issues (searching for relevant words or files should help you find your way around). We have an explanation of our [code structure](https://docs.pretalx.org/en/latest/developer/concepts.html) and the actual [data models](https://docs.pretalx.org/en/latest/developer/interfaces/models.html).

## How to do this

When a user switches to avatar retrieval from gravatar, retrieve that file (asynchronously would be fine, via a task, but you can start with direct retrieval, and we'll modify it later), and save it as speaker profile image.
(The client side request when the user switches to gravatar displays could be kept, for convenience, since it impacts only the one user deciding to opt in.) You'll have to change the three places where speaker profiles can be changed: `pretalx/orga/views/speaker.py` (by organisers), `pretalx/cfp/views/user.py` (by speakers), `pretalx/cfp/views/wizard.py` (by speakers during their initial submission). All of these should use the same form at `pretalx/person/forms.py`, though, so overriding that form's `save` method is the best way to go here. 

The scheduled `periodic_task` should check if the avatar has changed about once a day or weekly or so.

## Submit the PR

There are a number of tutorials on the Pull Request workflow online, for example this [short version](https://guides.github.com/introduction/flow/) by GitHub itself and this [more in-depth tutorial](https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github).

Remember to provide documentation and tests, if applicable. Run the tests before opening your PR, including style checks. We have a [contribution guide](https://docs.pretalx.org/en/latest/developer/contributing.html), which ought to prove helpful at this stage.

[...]

PR #1515, opened 2023-06-30
implement gravatar caching

fkusei:
This change implements local caching of gravatar profile images, thus avoiding possible GDPR issues.

If a person selects "use gravatar", celery will fetch the image from gravatar, store it locally, then disable gravatar import for that person. If the person wishes to refresh the avatar, they can select the gravatar checkbox again.

## How Has This Been Tested?
Local dev instance using production data

## Checklist

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -3,6 +3,7 @@
 Release Notes
 =============
 
+- :feature:`-` pretalx will now locally cache gravatar avatars to avoid GDPR issues when using gravatar.
 - :bug:`1498` Talks that were scheduled, but not confirmed by the speaker yet would be shown in the public speaker profile.
 - :feature:`-` pretalx always showed the anonymised version of proposals if there was one. Now it reverts to the non-anonymised one once the anonymisation period is over.
 - :feature:`-` Organiser pages for speakers now use their alphanumeric ``code`` identifier in the URL rather than the numeric ID, matching the public and API pages.
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -3,7 +3,7 @@
 Release Notes
 =============
 
-- :feature:`-` pretalx will now locally cache gravatar avatars to avoid GDPR issues when using gravatar.
+- :feature:`399` pretalx will now locally cache gravatar avatars to avoid GDPR issues when using gravatar.
 - :bug:`1498` Talks that were scheduled, but not confirmed by the speaker yet would be shown in the public speaker profile.
 - :feature:`-` pretalx always showed the anonymised version of proposals if there was one. Now it reverts to the non-anonymised one once the anonymisation period is over.
 - :feature:`-` Organiser pages for speakers now use their alphanumeric ``code`` identifier in the URL rather than the numeric ID, matching the public and API pages.
--- a/src/pretalx/person/forms.py
+++ b/src/pretalx/person/forms.py
@@ -24,6 +24,7 @@ from pretalx.common.mixins.forms import (
 )
 from pretalx.common.phrases import phrases
 from pretalx.person.models import SpeakerInformation, SpeakerProfile, User
+from pretalx.person.tasks import gravatar_cache
 from pretalx.schedule.forms import AvailabilitiesFormMixin
 from pretalx.submission.models import Question
 
@@ -237,6 +238,9 @@ class SpeakerProfileForm(
             else:
                 setattr(self.user, user_attribute, value)
             self.user.save(update_fields=[user_attribute])
+
+            if self.user.get_gravatar:
+                gravatar_cache.apply_async(args=(self.user.pk,))
 
         self.instance.event = self.event
         self.instance.user = self.user
--- a/src/pretalx/person/forms.py
+++ b/src/pretalx/person/forms.py
@@ -239,8 +239,8 @@ class SpeakerProfileForm(
                 setattr(self.user, user_attribute, value)
             self.user.save(update_fields=[user_attribute])
 
-            if self.user.get_gravatar:
-                gravatar_cache.apply_async(args=(self.user.pk,))
+        if self.user.get_gravatar:
+            gravatar_cache.apply_async(args=(self.user.pk,))
 
         self.instance.event = self.event
         self.instance.user = self.user
--- a/src/pretalx/person/models/user.py
+++ b/src/pretalx/person/models/user.py
@@ -275,23 +275,17 @@ class User(PermissionsMixin, GenerateCode, FileCleanupMixin, AbstractBaseUser):
 
     @cached_property
     def has_avatar(self) -> bool:
-        return self.get_gravatar or self.has_local_avatar
-
-    @cached_property
-    def has_local_avatar(self) -> bool:
         return self.avatar and self.avatar != "False"
 
     @cached_property
     def avatar_url(self) -> str:
-        if self.get_gravatar:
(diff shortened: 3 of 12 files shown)


src/pretalx/person/tasks.py
```diff
@@ -0,0 +1,40 @@
...
+from django.core.files.temp import NamedTemporaryFile
+from requests import get
+
+from pretalx.celery_app import app
+from pretalx.person.models.user import User
+
+logger = logging.getLogger(__name__)
+
+
+@app.task()
```
fkusei, line 12: I'm not sure what i'm doing wrong, but this task is never called by celery. It is discovered, but never called.

fkusei:
Not completely sure, should I add some clarification to the forms to tell people how gravatar integration works? I believe the current comment is accurate, but due to the new caching mechanism the fetched avatar is not immediately visible to the user, even when submitting the form (because submiting will only submit the task to the celery queue, but not wait for the result prior to reloading the page).

Alternatively, we could implement some JS magic to start the caching process immediately when selecting the option, then update the avatar image as soon as caching is done.

doc/changelog.rst
```diff
@@ -3,6 +3,7 @@
 Release Notes
 =============
 
+- :feature:`-` pretalx will now locally cache gravatar avatars to avoid GDPR issues when using gravatar.
```
rixx, line 6: Could link to #399 here!

src/pretalx/person/forms.py
```diff
@@ -238,6 +239,9 @@ def save(self, **kwargs):
                 setattr(self.user, user_attribute, value)
             self.user.save(update_fields=[user_attribute])
 
+            if self.user.get_gravatar:
+                gravatar_cache.apply_async(args=(self.user.pk,))
```
rixx, line 243: Shouldn't this be dedented, and not run once for every attribute in `self.user_fields`?

Could you also check a few lines above if we still need the `get_gravatar` handling there at all?

src/pretalx/person/models/user.py
```diff
@@ -275,23 +275,17 @@ def gravatar_parameter(self) -> str:
...
-        if self.has_local_avatar:
+        if self.has_avatar:
             return self.avatar.url
 
     def get_avatar_url(self, event=None):
         """Returns the full avatar URL, where user.avatar_url returns the
         absolute URL."""
-        if not self.avatar_url or "gravatar" in self.avatar_url:
+        if not self.avatar_url:
             return self.avatar_url
```
rixx, line 289: Please change this to `return ""`, as we can now be specific as to what we're returning.

src/pretalx/person/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+from pretalx.celery_app import app
+from pretalx.common.signals import periodic_task
+from pretalx.person.models.user import User
+
+logger = logging.getLogger(__name__)
+
+
+@app.task()
+def gravatar_cache(person_id: int):
+    user = User.objects.filter(pk=person_id).first()
```
rixx, line 17: `first()` is good, but then please add an early out in case this didn't return a user.

src/pretalx/person/tasks.py
```diff
@@ -0,0 +1,53 @@
...
+
```
rixx, line 18: Due to race conditions, I think it would be good to filter by `get_gravatar=True` here, or to check that it's still `True`.

(Consider the situation when updating pretalx to run this code: `runperiodic` runs every 1-5 minutes, but a larger instance would have more gravatar users than could be fetched in that time, resulting in duplicate fetches due to `refetch` being triggered again while one set of `refetch` is already running.)

rixx (changes requested):
Thank you so much! This has been a long-standing annoyance in pretalx, and I'm very much looking forward to having this in pretalx. I left some comments – would you prefer to continue to work on it, or should I take over?

src/pretalx/person/forms.py
fkusei: Oh yes, that should have been dedented.

About the gravatar handling above ... i'm not 100% sure, but i believe the field is either sent (if checked) or not sent at all when submitting the form, thus we explicitely need to set the value to `False` to unset gravatar.
Due to the immediate triggering of the caching it is highly unlikely to ever get manually un-set, but still possible (for example, if gravatar returns some error code and the user decides they want to manually upload the avatar instead)

src/pretalx/static/cfp/js/profile.js
```diff
@@ -34,23 +34,6 @@ $(function () {
...
-            $image.attr('src', "https://www.gravatar.com/avatar/" + $image.data('gravatar') + '?s=512');
-            $resetCheckbox.prop('checked', true);
-        } else if ($image.data('avatar')) {
-            $image.attr('src', $image.data('avatar'));
-            $image.removeClass('d-none');
-            $resetCheckbox.prop('checked', false);
-        } else {
-            $image.addClass('d-none');
-        }
-    });
```
fkusei, line 52: Please give me a hint here: Do we want to keep this code, so the submitting user is able to see which avatar will get cached?

fkusei: ... but then again we'd need all that CSP stuff again, don't we?

rixx: Yeah, I'd be in favour of keeping it. Most people won't know/remember/understand what "use gravatar" means, so having a preview is pretty much mandatory.

We'd only need the CSP stuff for these two views (the user's profile page and the profile step in the CfP process, which is still much better than enabling it on all schedule/talk/speaker pages.

Kunsi: nah, what i did is only halfway right, let me fix that.

Kunsi: Bugs fixed. All views where it's possible to select the gravatar checkbox will now load the image directly from gravatar if the checkbox is checked. The gravatar image is not auto-loaded on page load though, to prevent accidential leaking of ip addresses (especially relevant for orga people if the caching mechanism does not work for any reason).

rixx (approved):

rixx:
Thank you! I'll fix tests and code style after merging

src/pretalx/person/tasks.py
```diff
@@ -0,0 +1,60 @@
...
+from pretalx.celery_app import app
+from pretalx.common.signals import periodic_task
+from pretalx.person.models.user import User
+
+logger = logging.getLogger(__name__)
+
+
+@app.task()
+def gravatar_cache(person_id: int):
+    user = User.objects.filter(pk=person_id, get_gravatar=True).first()
```
fkusei, line 17: Seems like this filter does not work correctly for the task that's being generated during the profile save step.

I'm always getting `gravatar_cache() was called for user 856, but user was not found or user has gravatar disabled`, but it gets fetched just fine using the periodic task. Not sure what's wrong here though.

rixx: hmm, might be generated in a transaction, and the celery task is called too fast for the transaction to wrap up? I moved the whole logic to the model in 8669db20, as that's more future-safe for when we introduce a writable API. Can you check if that works?

fkusei: hm, now it doesn't update the avatar at all, even in the periodic task, even though the database field is `true`? I'm really confused.

rixx: Grr, that was too early in the morning. Yeah, that does break. 58547479 fixes.

PR #1530, opened 2023-07-24
fix: add fallback value for room guid

saerdnaer:
This PR resolves issues of #1338 by computing a fallback uuid from local ID + INSTANCE_IDENTIFIER, when the guid field in the room model is not filled with an external uuid

## How Has This Been Tested?

* run dev instance locally and checked
  * http://127.0.0.1:8000/democon/schedule/export/schedule.json
  * http://127.0.0.1:8000/democon/schedule/export/schedule.xml
  * http://127.0.0.1:8000/api/events/democon/rooms/ 
* run tests 

the existing guid field in the room form displays the fallback uuid as placeholder:
image

## Checklist

### DIFF
--- a/src/pretalx/agenda/templates/agenda/schedule.xml
+++ b/src/pretalx/agenda/templates/agenda/schedule.xml
@@ -19,7 +19,7 @@
         {% endfor %}
     </conference>
     {% for day in data %}<day index='{{ day.index }}' date='{{ day.start.date|date:"c" }}' start='{{ day.start|date:"c" }}' end='{{ day.end|date:"c" }}'>
-        {% for room in day.rooms %}<room name='{{ room.name|xmlescape }}' guid='{{ room.uuid() }}'>
+        {% for room in day.rooms %}<room name='{{ room.name|xmlescape }}' guid='{{ room.guid }}'>
             {% for talk in room.talks %}<event guid='{{ talk.uuid }}' id='{{ talk.submission.id }}'>
                 <room>{{ room.name|xmlescape }}</room>
                 <title>{{ talk.submission.title|xmlescape }}</title>
--- a/src/pretalx/agenda/templates/agenda/schedule.xml
+++ b/src/pretalx/agenda/templates/agenda/schedule.xml
@@ -19,7 +19,7 @@
         {% endfor %}
     </conference>
     {% for day in data %}<day index='{{ day.index }}' date='{{ day.start.date|date:"c" }}' start='{{ day.start|date:"c" }}' end='{{ day.end|date:"c" }}'>
-        {% for room in day.rooms %}<room name='{{ room.name|xmlescape }}'{% if room.guid %} guid='{{ room.guid }}'{% endif %}>
+        {% for room in day.rooms %}<room name='{{ room.name|xmlescape }}' guid='{{ room.uuid() }}'>
             {% for talk in room.talks %}<event guid='{{ talk.uuid }}' id='{{ talk.submission.id }}'>
                 <room>{{ room.name|xmlescape }}</room>
                 <title>{{ talk.submission.title|xmlescape }}</title>
--- a/src/pretalx/api/serializers/room.py
+++ b/src/pretalx/api/serializers/room.py
@@ -1,5 +1,5 @@
 from i18nfield.rest_framework import I18nAwareModelSerializer
-from rest_framework.serializers import ModelSerializer, SerializerMethodField
+from rest_framework.serializers import CharField, ModelSerializer, SerializerMethodField
 
 from pretalx.schedule.models import Availability, Room
 
@@ -17,6 +17,7 @@ class AvailabilitySerializer(ModelSerializer):
 
 class RoomSerializer(I18nAwareModelSerializer):
     url = SerializerMethodField()
+    guid = CharField(source="uuid")
 
     def get_url(self, obj):
         return obj.urls.edit
@@ -25,7 +26,7 @@ class RoomSerializer(I18nAwareModelSerializer):
         model = Room
         fields = (
             "id",
-            "uuid",
+            "guid",
             "name",
             "description",
             "capacity",
--- a/src/pretalx/api/serializers/room.py
+++ b/src/pretalx/api/serializers/room.py
@@ -25,7 +25,6 @@ class RoomSerializer(I18nAwareModelSerializer):
         model = Room
         fields = (
             "id",
-            "guid",
             "uuid",
             "name",
             "description",
--- a/src/pretalx/api/serializers/room.py
+++ b/src/pretalx/api/serializers/room.py
@@ -23,7 +23,16 @@ class RoomSerializer(I18nAwareModelSerializer):
 
     class Meta:
         model = Room
-        fields = ("id", "guid", "name", "description", "capacity", "position", "url")
+        fields = (
+            "id",
+            "guid",
+            "uuid",
+            "name",
+            "description",
+            "capacity",
+            "position",
+            "url",
+        )
 
 
 class RoomOrgaSerializer(RoomSerializer):
--- a/src/pretalx/locale/de_DE/LC_MESSAGES/django.po
+++ b/src/pretalx/locale/de_DE/LC_MESSAGES/django.po
@@ -7929,11 +7929,13 @@ msgstr ""
 msgid "GUID"
 msgstr "GUID"
 
-#: pretalx/schedule/models/room.py:25
-msgid "Unique identifier (UUID) to help external tools identify the room."
-msgstr ""
-"Eindeutiger Bezeichner (UUID), der externen Anwendungen hilft, diesen Raum "
-"zu identifizieren."
+#: pretalx/schedule/models/room.py:32
+msgid ""
+"Unique identifier (UUID) to help external tools identify the room. When left "
+"empty, a UUID will be generated automatically from the pretalx instance identifier "
+"and local room id."
+msgstr "Eindeutiger Bezeichner (UUID), der externen Anwendungen hilft diesen Raum "
+"zu identifizieren. Wenn leer, wird der Wert aus der pretalx Instanz UUID und der lokalen Raumnummer berechnet "
 
 #: pretalx/schedule/models/room.py:33
 msgid "A description for attendees, for example directions."
--- a/src/pretalx/locale/de_DE/LC_MESSAGES/django.po
+++ b/src/pretalx/locale/de_DE/LC_MESSAGES/django.po
@@ -7930,12 +7930,10 @@ msgid "GUID"
 msgstr "GUID"
 
 #: pretalx/schedule/models/room.py:32
-msgid ""
-"Unique identifier (UUID) to help external tools identify the room. When left "
-"empty, a UUID will be generated automatically from the pretalx instance identifier "
-"and local room id."
-msgstr "Eindeutiger Bezeichner (UUID), der externen Anwendungen hilft diesen Raum "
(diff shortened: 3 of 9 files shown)


src/pretalx/schedule/models/room.py
```diff
@@ -71,3 +73,16 @@ class urls(EventUrls):
...
+        """Either a UUID5 calculated from the submission code and the instance identifier; 
+        or GUID value of the room, if it was imported or set manually."""
+        if self.guid:
+            return self.guid
+
+        global INSTANCE_IDENTIFIER
+        if not INSTANCE_IDENTIFIER:
+            from pretalx.common.models.settings import GlobalSettings
+
+            INSTANCE_IDENTIFIER = GlobalSettings().get_instance_identifier()
```
saerdnaer, line 87: What is the proper way to access the `INSTANCE_IDENTIFIER`?

rixx (commented):
This value should also be

- visible in the room form
- included in the API
- used in the base exporter data

src/pretalx/agenda/templates/agenda/schedule.xml
```diff
@@ -14,7 +14,7 @@
         <time_zone_name>{{ event.timezone }}</time_zone_name>
     </conference>
     {% for day in data %}<day index='{{ day.index }}' date='{{ day.start.date|date:"c" }}' start='{{ day.start|date:"c" }}' end='{{ day.end|date:"c" }}'>
-        {% for room in day.rooms %}<room name='{{ room.name|xmlescape }}'{% if room.guid %} guid='{{ room.guid }}'{% endif %}>
+        {% for room in day.rooms %}<room name='{{ room.name|xmlescape }}' guid='{{ room.guid }}'>
```
rixx, line 17: This is supposed to be `room.uuid`, isn't it?

src/pretalx/api/serializers/room.py
```diff
@@ -23,7 +23,16 @@ def get_url(self, obj):
 
     class Meta:
         model = Room
-        fields = ("id", "guid", "name", "description", "capacity", "position", "url")
+        fields = (
+            "id",
+            "guid",
```
rixx, line 28: I don't think we need `guid` and `uuid`, just `uuid` should be enough, right? We never want `uuid` when `guid` is set, after all, as that would be confusing.

src/pretalx/locale/de_DE/LC_MESSAGES/django.po
```diff
@@ -4367,7 +4367,7 @@ msgstr "Reviewer einteilen"
 msgid "Export reviews"
 msgstr "für Reviewerinnen"
 
-#: pretalx/orga/templates/orga/base.html:336 pretalx/schedule/models/room.py:40
```
rixx, line 4370: Please don't commit changes like this: they don't carry any meaning and will lead to merge conflicts down the line.

rixx (commented):
Some more comments, shaping up really well!

src/pretalx/api/serializers/room.py
saerdnaer: It depends on your use case... I explicitly included both so one can debug if the guid was filled in manually or automatically generated – but we can change that if you prefer to only have one.

src/pretalx/schedule/forms.py
```diff
@@ -248,6 +248,7 @@ def __init__(self, *args, **kwargs):
             "Information for speakers, e.g.: Projector has only HDMI input."
         )
         self.fields["capacity"].widget.attrs["placeholder"] = "300"
+        self.fields["guid"].widget.attrs["placeholder"] = self.instance.uuid or ""
```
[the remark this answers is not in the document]
saerdnaer: The idea was to be able to differentiate if the guid was filled in externally or not, thereby I currently only fill the guid field in the database if the user explicitly entered it.

src/pretalx/agenda/templates/agenda/schedule.xml
saerdnaer: No, the templates only reads the dicts created in schedule/exports.py – its enough to change the mapping there: 
image

src/tests/agenda/test_agenda_schedule_export.py
```diff
@@ -106,7 +106,7 @@ def test_schedule_frab_json_export(
     django_assert_max_num_queries,
     orga_user,
 ):
-    with django_assert_max_num_queries(12):
+    with django_assert_max_num_queries(14):
```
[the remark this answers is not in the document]
saerdnaer: did a rebase onto current main, the numbers did not change

src/pretalx/api/serializers/room.py
rixx: Yes, please. I want this feature, in use for only a very small group of users, to take up as little space and maintenance effort as possible (to the point that I keep regretting that we didn't put it into a plugin.)

src/pretalx/schedule/forms.py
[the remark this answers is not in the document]
rixx: I can see the intention, but a) I don't think it really makes a difference, and b) setting the placeholder might be the most frustrating solution in that case, because it will *show* the generated UUID to the user, but it won't allow the user to copy it, thereby forcing them to either copy it manually or head into the page source code in order to retrieve it.

src/tests/agenda/test_agenda_schedule_export.py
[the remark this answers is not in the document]
rixx: Can you confirm that the additional queries are the ones retrieving the instance identifier?

src/pretalx/agenda/templates/agenda/schedule.xml
rixx: Oh wow, this will absolutely 100% bite somebody during debugging at some point lol.

I hate it, but I guess there's nothing to be done, as it's all just part of the feature.

src/pretalx/schedule/models/room.py
```diff
@@ -23,7 +27,7 @@ class Room(LogMixin, models.Model):
         blank=True,
         verbose_name=_("GUID"),
         help_text=_(
-            "Unique identifier (UUID) to help external tools identify the room."
+            "Unique identifier (UUID) to help external tools identify the room. When left empty, a UUID will be generated automatically from the pretalx instance identifier and local room id."
```
rixx, line 30: That's a very technical way of putting things, which is not really suitable for users (who already end up confused about the difference in naming between "GUID" and "UUID", which is why we've had to hide that part of the form!). Not sure we need to change anything here at all.

saerdnaer: So as it's already hidden by default, the addition is is okay? Or should I drop the addtional text from this PR?

src/pretalx/schedule/forms.py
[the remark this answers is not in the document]
saerdnaer: Endusers never should need to copy the guid from the form field: 

* You either import rooms with their guid via schedule.json/xml – and if someone change their name the downstream system can still identify the smae room by its guid. 
* The other case is when room was already assigned an ID in another system, then you can set the guid field to that existing value.

src/tests/agenda/test_agenda_schedule_export.py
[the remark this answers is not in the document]
saerdnaer: what is the proper way to debug/confirm this?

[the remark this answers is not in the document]
saerdnaer: nevermind, this issue resolved itself.

src/pretalx/api/serializers/room.py
```diff
@@ -23,7 +23,15 @@ def get_url(self, obj):
 
     class Meta:
         model = Room
-        fields = ("id", "guid", "name", "description", "capacity", "position", "url")
+        fields = (
+            "id",
+            "uuid",
```
rixx, line 28: This changes the REST API to provide a "uuid" attribute rather than a "guid" attribute. That's a breaking change, which doesn't seem necessary, right?

src/pretalx/schedule/models/room.py
rixx: I'd favour dropping the text, as it doesn't add much in understanding,and will force every translation to add the new sentence (and otherwise the whole text will display in English). As we don't have lots of active translators, I'd like to avoid unnecessary modifications to existing strings.

rixx (commented):
Apart from the API change, this is probably as good as it gets with this approach!

I'm still not entirely clear why we don't just generate a GUID on create (while allowing the user to change it later, if they want), as we initially discussed when you added this feature. Is there any big advantage to having user-set and auto-generated GUIDs distinct like this?

src/pretalx/api/serializers/room.py
saerdnaer: oh, it totally missed that... what is the proper way to export the value of the uuid property with the guid key?

rixx: Might work with a `UUIDField(source='uuid')`, but I'm not sure if `source` doesn't maybe check for actual proper fields. In that case, a `SerializerMethodField()`, and `get_guid(self, instance): return instance.uuid`, -ish, off the top of my head.

src/pretalx/schedule/forms.py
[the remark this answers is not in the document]
rixx: "Should" never need to copy it, yes. But if the data is there, the user "should" be able to retrieve it by a different way than downloading a weird XML export, or copying by sight, or going to browser tools.

I still feel like this would not be a problem if we just always generated the `guid` (but allowed the user to change it). Am I overlooking a critically useful aspect of always computing it when it's not explicitly chosen by the user?

rixx:
> I'm still not entirely clear why we don't just generate a GUID on create (while allowing the user to change it later, if they want), as we initially discussed when you added this feature. Is there any big advantage to having user-set and auto-generated GUIDs distinct like this?

This is my last open question, fwiw.

rixx:
Ping?

saerdnaer:
Ah, sorry for the delay, I wanted to be able to differentiate between the "somebody added an external Id" and "we did not know the external ID when creating the room" to reduce multiple guid's for the same entity

rixx:
You'll probably want to rebase this PR to see if tests are passing now.

Kunsi:
Hopefully the rebase worked correctly.

rixx:
Looks like you committed a merge conflict.

Kunsi:
Do you see where? Because i can merge that fine locally:
```
/home/kunsi/git/pretalx/src|main✓➤ git pull upstream main
From https://github.com/pretalx/pretalx
 * branch              main       -> FETCH_HEAD
Updating 0cd1d6b2..d3eaea0b
Fast-forward
 src/pretalx/common/models/settings.py | 10 +++++-----
 1 file changed, 5 insertions(+), 5 deletions(-)

/home/kunsi/git/pretalx/src|main✓➤ git merge room-guid-fallback
Merge made by the 'ort' strategy.
 src/pretalx/agenda/templates/agenda/schedule.xml |  2 +-
 src/pretalx/api/serializers/room.py              | 13 +++++++++++--
 src/pretalx/locale/de_DE/LC_MESSAGES/django.po   |  2 +-
 src/pretalx/locale/django.pot                    |  2 +-
 src/pretalx/schedule/exporters.py                |  4 ++--
 src/pretalx/schedule/forms.py                    |  1 +
 src/pretalx/schedule/models/room.py              | 20 ++++++++++++++++++++
 7 files changed, 37 insertions(+), 7 deletions(-)
```

Kunsi:
Whoops, thanks for pointing that out!

rixx (approved):
Probaby fastest if I do the remaining cleanup myself after the merge.

PR #1543, opened 2023-08-05
Allow to discard draft proposals

### LINKED ISSUE #1374: Fix remaining draft proposal issues
Draft proposals as per #672 have been implemented in #1373, but there are a few remaining sub-features that need to be implemented before we can release. (The semi-done feature was merged to facilitate early testing).

- [x] **Send emails to users with drafts.** These emails are intended as reminders sent by the organisers. The tricky bit is that organisers are not meant to know who has unsubmitted drafts lying around, which has a bunch of unfortunate implications:
	- [x] Drafts cannot be used to generate email previews
	- [x] Emails to users with drafts have to either be sent immediately, **or** be hidden in the outbox (not sure which is worse)
	- [x] Emails to users with drafts must not show up in any history, including the action log and the list of emails shown on a user's profile. We'll probably want a `QueuedMail.private` or `QueuedMail.hidden` flag here
- [ ] Discard drafts: users should be allowed to discard/remove drafts, including after the CfP has closed
- [ ] Currently, mandatory profile information and questions can remain unanswered. We need to somehow check for those, and require the user to update their profile page before returning to finalise the draft.

rixx:
Closes #1374, as the remaining issue is better off in a separate issue

### DIFF
--- a/src/pretalx/cfp/templates/cfp/event/user_submission_discard.html
+++ b/src/pretalx/cfp/templates/cfp/event/user_submission_discard.html
@@ -1,0 +1,35 @@
+{% extends "cfp/event/base.html" %}
+{% load bootstrap4 %}
+{% load compress %}
+{% load i18n %}
+{% load static %}
+
+{% block title %}{% translate "Discard draft proposal" %} :: {% endblock %}
+
+{% block content %}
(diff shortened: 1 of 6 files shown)


rixx (commented):
Even added tests

PR #1575, opened 2023-09-10
docs: package root not under src

johanvdw:
Fixes the dev instructions, you should no longer enter src before installing.


rixx (approved):
Thanks!

PR #1579, opened 2023-09-13
fix: commands/init.py

lorenzleutgeb:
I think that `pretalx init` is broken in the `v2023.1.0` release. There's a syntax error, which I fix in this PR, and the behavior of `--noinput` was also broken. I copied it from `createsuperuser` and now it works for me.

## Checklist

### DIFF
--- a/src/pretalx/common/management/commands/init.py
+++ b/src/pretalx/common/management/commands/init.py
@@ -33,7 +33,8 @@ class Command(BaseCommand):  # pragma: no cover
         parser.add_argument(
             "--noinput",
             "--no-input",
-            action="store_true",
+            action="store_false",
+            dest="interactive",
             help="Suppresses all interactive prompts and uses environment variables instead. "
             "If environment variables are missing, the command will fail. "
             "Required environment variables:\n"
@@ -60,7 +61,8 @@ class Command(BaseCommand):  # pragma: no cover
             """\nLet\'s get you a user with the right to create new events and access every event on this pretalx instance."""
         )
 
-        call_command("createsuperuser", no_input=options["noinput"])
+        call_command("createsuperuser", interactive=options["interactive"])
+
         user = User.objects.order_by("-id").filter(is_administrator=True).first()
 
         self.stdout.write(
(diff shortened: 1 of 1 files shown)


lorenzleutgeb:
If I understand correctly, the failed test is expected to raise `CommandError`, but instead it raises `OSError`. Any hints?

src/pretalx/common/management/commands/init.py
```diff
@@ -90,8 +92,8 @@ def handle(self, *args, **options):
             ),
         )
         self.stdout.write(self.style.SUCCESS("\nNow that this is done, you can:"))
-        self.stdout.write(" - Create your first event at {event_url}").format(
-            event_url=event_url
+        self.stdout.write(
+            " - Create your first event at {event_url}".format(event_url=event_url)
```
rixx, line 96: I think all of these could be f-strings now that they aren't translated anymore, I'll take care of that, too

rixx (approved):
`init` got a change just before the release, and I could've sworn there were both automated and manual tests, but apparently I didn't test carefully enough (or broke things when rebasing …).

Thank you in any case, much appreciated! (And sorry that you seem to run into so many problems with pretalx and its setup atm!)

rixx:
Merging, as IIRC the switch from `CommandError` to `OSError` was part of the merged feature that added the `--no-input` flag. I assumed that the switch was due to the changed input mode, but apparently it indicated a real problem. I'll switch the tests back to the old OSError.

lorenzleutgeb:
Thanks a lot for merging! Would you please consider cherry-picking this PR (and possible other fixes) on top of the 2023.1.0 release and create a new release, like 2023.1.1? If I am not mistaken, the syntax error in `init.py` would prevent people from successfully initializing, thus the 2023.1.0 release has severe issues. In [ngi-nix/ngipkgs we opted to package 2023.1.0, but added this PR as a patch](https://github.com/ngi-nix/ngipkgs/blob/173038063030534edece2d0cc8fbebb2564ba3ed/pkgs/pretalx/default.nix#L41-L46).

rixx:
There is at least one bug in 2023.1.0 still at large, which I would like to fix before moving onto a patch release.

rixx:
Found the remaining bugs (turned out to be more than one, sigh), and this change is now part of [v2023.1.3](https://pretalx.com/p/news/releasing-pretalx-2023-1-3/).

PR #1589, opened 2023-09-27
Use django-table2

rixx:
This PR attempts to include django-table2 in pretalx.

## Status

The proposal list has been replaced with a feature-complete equivalent table using django-table-2, except that the changed code does not as yet respect pagination page size from urlparams. Queries have increased by a constant amount to permit links to submission types (which we didn't have previously).

## Performance

Rendering tables with django-table2 seems to be noticeably slower than running through our previous giant template. More so in `DEBUG` mode (as templates aren't cached then, and we're using more templates), but even in `DEBUG=False`, there seems to be a significant penalty:

```
## debug=true
page size  old        django-tables2
25         550-700ms   800-950ms
100        750-950ms   900-1200ms


## debug=false
page size  old        django-tables2
25         250-330ms   300-350ms
100        350-470ms   580-700ms
```

I tried to attach py-spy, and can't spot anything obvious in the block belonging to `django-table2` (svg attached). 
2023-09-27T17:14:16+02:00 

<details>
<summary>SVG behind the fold</summary>

profile

</details>

With a performance penalty like this, I'm against using `django-table2` despite the obvious benefits in maintainable code, especially as larger pages seem to be hit disproportionately, and our review dashboard is an *unpaginated* page of all proposals. (Of course, once we have decent bulk editing, we can paginate or lazy-load there, it's not like we're out of solutions, but still).

I still want to see perf penalties on the big review dashboard, and then I'll decide what to do.

## Todo

- [ ] Decide if we want to use django-table2
- [ ] Migrate all lists to django-table2
    - [ ] review dashboard
    - [ ] questions list
    - [ ] feedback list
    - [ ] allfeedback list
    - [ ] taglist
    - [ ] speakerlist
    - [ ] informationlist
    - [ ] eventhistory
    - [ ] submissiontypelist
    - [ ] tracklist
    - [ ] accesscodelist
- [ ] Build generic filter form UI
- [ ] Build table builder / column selection UI
- [ ] Build exporter UI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dependencies = [
   "django-i18nfield~=1.9.0",
   "django-libsass~=0.8",
   "django-scopes~=2.0.0",
+  "django-tables2~=2.6.0",
   "djangorestframework~=3.14.0",
   "libsass~=0.22.0",
   "Markdown~=3.4.0",  # https://python-markdown.github.io/change_log/
--- a/src/pretalx/orga/tables/submission.py
+++ b/src/pretalx/orga/tables/submission.py
@@ -1,0 +1,212 @@
+from urllib.parse import quote
+
+import django_tables2 as tables
+from django.template import Context
+from django.utils.safestring import mark_safe
+from django.utils.translation import gettext_lazy as _
+
+from pretalx.submission.models import Submission
+
+
+def get_header(request):
+    return "hey"
+
+
+class ContextTemplateColumn(tables.TemplateColumn):
+    """Allow to change the context_object_name."""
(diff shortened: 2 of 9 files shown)


rixx:
There's of course the option to take the performance hit in return for easier maintenance, but I think due to the very low amount of features I'm likely to use, I'd rather look at some of the `django-table2` code and build my own version (with *much* less bells and whistles, mostly some iterators etc), which could also integrate better with what pretalx needs – you'll see that I had to already do some pretty gnarly subclassing because `django-table2` doesn't really believe in making defaults easy to override …

Issue #1594, opened 2023-10-03
Plugin hook for sent email


Kunsi:
### Problem and impact

For tracking communication with speakers it would be really nice if sent emails would get imported automatically into some sort of ticket system. Due to the many ticket systems available, i don't think it is suitable to add such functionality directly into pretalx.

Instead, pretalx should have a plugin hook so plugins can get notified of sent emails, and then take care of importing those emails into the ticket system of choice.

### Proposed solution / feature

In my opinion, it would be best to implement this somewhere in `common.mail.mail_send_task()`, because that's the only place (i could find) that knows of all mails and is running in celery.

rixx:
Definitely a cool idea!

I don't think we want to use `mail_send_task()` though – this is used for non-event mails as well as things like password resets and automated confirmation emails ("we have received your proposal"). I think putting something into `QueuedMail.send()` however sounds sensible. We'd need to check and see if we want to call this only if the `QueuedMail` object has an `id` (aka if it's still accessible for users after the fact), but we could also let the plugin handle this check.

As signals run synchronously, it'd probably be **very highly recommended** for any receiving signal to call a task asynchronously (as not to block email sending), but other than that, I don't really see any problems with this.

PR #1595, opened 2023-10-03
Add plugin hook for sent `QueuedMail`s

Kunsi:
This implements #1594

## How has this been tested?
- local dev instance of pretalx

## Checklist

### DIFF
--- a/src/pretalx/mail/models.py
+++ b/src/pretalx/mail/models.py
@@ -16,6 +16,7 @@ from pretalx.common.mixins.models import LogMixin
 from pretalx.common.templatetags.rich_text import ALLOWED_TAGS
 from pretalx.common.urls import EventUrls
 from pretalx.mail.context import get_mail_context
+from pretalx.mail.signals import queuedmail_has_been_sent
 
 
 class MailTemplate(LogMixin, models.Model):
@@ -322,6 +323,11 @@ class QueuedMail(LogMixin, models.Model):
         )
 
         self.sent = now()
+        queuedmail_has_been_sent.send(
+            sender=self.event,
+            mail=self,
+        )
+
         if self.pk:
             self.log_action(
                 "pretalx.mail.sent",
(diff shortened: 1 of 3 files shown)


src/pretalx/mail/models.py
```diff
@@ -322,6 +323,11 @@ def send(self, requestor=None, orga: bool = True):
         )
 
         self.sent = now()
+        queuedmail_has_been_sent.send(
```
rixx, line 326: Let's call the signal after the object has been saved, I'd say. Otherwise, this can easily lead to inconsistencies, and the signal name implies that the action has been completed.

rixx (commented):
Thank you!

I think I'd prefer a shorter signal name – either `queuedmail_sent`, or, anticipating the obvious future use case, `queueuedmail_post_send` (as I'm sure eventually we'll have a pre-send hook, too …). What do you think?

src/pretalx/mail/models.py
Kunsi: I agree, i was just confused by `self.save()` only getting called if `self.pk` evaluates to true. That's why this is here. I will move it to after the `if` clause.

rixx: The idea is basically to allow the use of `QueuedMail` objects as interfaces for email sending (generating pretty HTML, event settings etc) but without adding those emails to the organiser-accessible log. This is e.g. used when sending emails to users with draft proposals, who should **not** be visible to organisers, ever!

Actually, due to this, we may want to only trigger the signal in the `if self.pk` block after all.

rixx: Reviewing pretalx code, this hidden sending is used

- to send organisers informations ("News from your content system")
- to send out access codes
- when one speaker invites other speakers to their proposal (also nothing organisers ought to see!)
- when sending a notification to the organiser address that there has been a new proposal


So I'd say the signal can just go in the `pk=True` branch of the code, where it won't accidentally read hidden information. If a plugin really wants to see **all** outgoing emails, the easiest way would be to implement its own email backend (but then it would have to handle these distinctions by itself, especially the personal messages from speakers to other people, and the draft proposals, where pretalx makes guarantees of not leaking the identities and contents of non-submitted proposals).

rixx (approved):

PR #1616, opened 2023-11-02
Expand timeslices in schedule editor on hover

rixx:
This PR is very WIP and does not fully work yet. Status: Timeslices can expand on hover, but they flicker and aren't clickable.

Idea:

- new state, `tentativeExpandedSlices` instead of `hoverSlices`, which can be made permanent with a click (adding those timeslices to the expanded slices)
- re flickering: rash says to watch [`hoverSlice`](https://github.com/pretalx/pretalx/blob/main/src/pretalx/frontend/schedule-editor/src/components/GridSchedule.vue#L61), but that's currently only updated when dragging, bc expensive. we could stuff that into `updateHoverSlice` (rename that function) and bail directly afterwards …


rixx:
Abandoning: this is super buggy unless you do some serious observing (`@mouseleave` doesn't really cut it in many cases), and even if it works perfectly, the constant change in schedule size is just janky and unpleasant. I put some other ideas into the tracking issue.

PR #1617, opened 2023-11-04
Adjust plugin urlexample to django 4

johanvdw:
The current documentation for writing a plugin seems out of date, as url is no longer present in Django 4.

Please check carefully, I've not tested all alternatives.

### DIFF
--- a/doc/developer/plugins/customview.rst
+++ b/doc/developer/plugins/customview.rst
@@ -18,18 +18,16 @@ Organiser panel views
 If you want to add a custom view to the organiser area of an event, register an URL in your
 ``urls.py`` that lives in the ``/orga/`` subpath::
 
-    from django.conf.urls import url
-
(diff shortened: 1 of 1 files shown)


rixx (changes requested):
The trouble with this is that it permits non-event-slugs in URL routing. However, we of course need to update the docs to reflect currently valid code – which in this case is `re_path`.

rixx (changes requested):
The trouble with this is that it permits non-event-slugs in URL routing. However, we of course need to update the docs to reflect currently valid code – which in this case is `re_path`.

rixx:
I'll make the change myself. I'll also add an `urls.py` to `pretalx-plugin-cookiecutter`, as I expect most plugins will want one.

(If you want to share the plugin you're building, btw, there is a [wiki page](https://github.com/pretalx/pretalx/wiki/Plugins) listing plugins!)

PR #1636, opened 2023-12-17
Make sure proposal can keep track if submission is closed

johanvdw:
We used the "require access token" to close certain tracks before the whole conference cfp was closed.

After this was done, submitters who wanted to update their proposal were unable to save it unless they also changed the track.

In any case, I think it is logical that you can keep the existing track if you edit a proposal.

## How has this been tested?

Manual test + deployed on live instance. Still need to add a unit test. If you have a suggestion of a similar test I'm willing to do so.

Existing unit tests fail, but that does not seem related.

## Checklist

### DIFF
--- a/src/pretalx/submission/forms/submission.py
+++ b/src/pretalx/submission/forms/submission.py
@@ -1,5 +1,5 @@
 from django import forms
-from django.db.models import Count, Exists, OuterRef
+from django.db.models import Count, Exists, OuterRef, Q
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 from django_scopes.forms import SafeModelChoiceField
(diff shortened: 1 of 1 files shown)


src/pretalx/submission/forms/submission.py
```diff
@@ -72,9 +72,14 @@ def _set_track(self, instance=None):
 
             access_code = self.access_code or getattr(instance, "access_code", None)
             if not access_code or not access_code.track:
-                self.fields["track"].queryset = self.event.tracks.filter(
+                filter = self.event.tracks.filter(
                     requires_access_code=False
                 )
+                # ensure current track selection can be preserved
+                if instance and instance.track:
```
rixx, line 79:
```suggestion
                if instance and instance.track and instance.track.requires_access_code:
```

don't really need to add that filter if the track isn't restricted, right?

rixx (commented):
yeah, makes total sense. Just one small niggle ^

src/pretalx/submission/forms/submission.py
```diff
@@ -72,9 +72,14 @@ def _set_track(self, instance=None):
 
             access_code = self.access_code or getattr(instance, "access_code", None)
             if not access_code or not access_code.track:
-                self.fields["track"].queryset = self.event.tracks.filter(
+                filter = self.event.tracks.filter(
```
rixx, line 75: Please rename this to `_filter` or `track_filter` – `filter` is a built-in method in Python, and even though we don't use it here in this method, we don't want to start overriding built-ins 😉

rixx (approved):
Great, thank you!

PR #1645, opened 2023-12-29
use speaker code consistently in warnings

johanvdw:
It's a bit inconsistent to return the speaker code for an availability warning (line 422) and the id for an overlap warning.

somewhat related: it may be a good idea to use different types for both warnings.


rixx (approved):
Hm, I was worried that this would break the schedule editor, but apparently we don't use the extra metadata currently at all. Should probably do that 😬

PR #1538, opened 2023-07-26
schedule.json extensions

saerdnaer:
This PR adds a few missing attributes to schedule.json export, including speaker avatars and switching from relative URIs for external assets to absolute URLs. 

## How Has This Been Tested?

* run locally with democon test data (which has no images on event nor speaker/person)
* validated output agains schedule.json json-schema

## Screenshots (if appropriate):
image

## Checklist

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] My change is listed in the CHANGELOG.rst if appropriate.

PR #1676, opened 2024-02-10
pretalx.common.models.settings: correctly set INSTANCE_IDENTIFIER if …

Kunsi:
I believe this is what f77dbce1dabc4098c2fad449096de754bab7f47f intended to do, but was missed.

Found this while revamping my dev setup to do work on #1538


rixx (commented):
"mishap" is a bit weird in terms of wording, but I'll just fix that post-merge. Thanks.

rixx (commented):
(Actually, looking at this some more, I shouldn't have merged it – I'll rework it completely. Thank you for pointing out the bug, though.)

Kunsi:
Ah yes, your code looks much better than my clunky implementation :)

PR #1689, opened 2024-02-18
removed superfluous colons in German translations on speaker submission page

pascoda:
Currently, e.g. on https://fahrplan.eh21.easterhegg.eu/ submissions, there are double colons visible after "Vortragende" (::).

That's because both the translation and the code that creates the page contain colons.

Might be a wider issue: the msgid includes colons, too, but not all (other) translations do. E.g., one of the variations in the arabic translations includes colons, the rest doesn't.


rixx (changes requested):
Thanks for the bug report – however, this isn't how translations issues like this can be fixed. As the problem is there in the original English string, it has to be adjusted there, and then in all the translations. I'll take care of that.

In the future, please use translate.pretalx.com to edit translations, as that can take place of possible merge conflicts.

Issue #1730, opened 2024-04-05
allow scheduling of "pending accepted" talks


margau:
### Problem and impact

When i'm planning an event/designing a schedule, I don't want to finally already accept a talks (so that speakers could see the accepted state after login), but already try out how everything fits in the schedule editor. Currently, pending accepted proposals don't show up in the schedule editor.

### Proposed solution / feature

We could make pending accepted talks show up in the schedule editor, or make this configurable.
This would however require the availability of "Start" and "End" time, which is not currently available for talks in "pending accepted" state.
Maybe there is another option?

rixx:
Thank you again for requesting and building this!

PR #1732, opened 2024-04-05
feat (#1730): Allows scheduling of pending accepted/pending confirmed talks

margau:
This PR is a draft implementation for #1730, and allows scheduling (by creating slots) for talks that are not finally accepted/confirmed, but only pending.

This allows internal time planning for organizers while not yet having to confirm the talk.
Public visibility is not changed.

## How has this been tested?

Manually tested it by moving the states forward & backward
Bildschirmfoto vom 2024-04-05 22-25-59
Bildschirmfoto vom 2024-04-05 22-26-18

## Checklist

## Open Points/RFC

This PR, at the current stage, is intended as an RFC. Open Points:
- in general, would you merge this feature? If yes, I'd clean/finish the PR.
- shall we add a config option to enable this behavior? On or off by default?
- I'm not so happy with the calls of `update_talk_slots()` - maybe this should be called at the end of every Submission Update to reduce complexity?

### DIFF
--- a/src/pretalx/frontend/schedule-editor/src/components/Session.vue
+++ b/src/pretalx/frontend/schedule-editor/src/components/Session.vue
@@ -8,6 +8,9 @@
 	.info
 		.title {{ getLocalizedString(session.title) }}
 		.speakers(v-if="session.speakers") {{ session.speakers.map(s => s.name).join(', ') }}
+		.pending-line(v-if="session.state && session.state !== 'confirmed' && session.state !== 'accepted'") 
+			i.fa.fa-exclamation-circle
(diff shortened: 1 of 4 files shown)


margau:
I've pushed a version (dotted border, clear text indicating the pending state) for all non-accepted or non-confirmed talks:
image

This is not relevant for talks already accepted but pending confirmation, which are shown as normal accepted talks.

What do you think about this visualization?

rixx (approved):
I'll make some small adjustments after the merge, but nothing to worry about. Thank you so much for your contribution!

PR #1746, opened 2024-04-16
Don't recommend `sudo` in a venv

strugee:
Using `sudo` in a venv is a) not necessary and b) potentially dangerous since it may cause system `pip3` to be used.

## How has this been tested?

I tested the fixed command, albeit on a Fedora machine.

## Checklist

### DIFF
--- a/doc/developer/setup.rst
+++ b/doc/developer/setup.rst
@@ -54,7 +54,7 @@ do so automatically).
 If you are working on Ubuntu or Debian, we strongly recommend upgrading your pip and setuptools
 installation inside the virtual environment, otherwise some of the dependencies might fail::
 
-    sudo pip3 install -U pip setuptools wheel
+    pip3 install -U pip setuptools wheel
 
 
 Get a copy of the source code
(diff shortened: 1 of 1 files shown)


strugee:
Saw a related point of potential confusion that this patch made worse, so I fixed that too.

rixx (approved):
Thanks! That's a good point and is a leftover from ages ago. I changed the first line myself as I don't believe running the `git` and `cd` command within a venv is mandatory, so only that part of the change is necessary imo.

rixx:
(Just to address it for future reference: I wouldn't agree with the characterisation as "potentially dangerous", as updating `setuptools` and `wheel` with global pip shouldn't be an issue – it's certainly not what the command is intended to do here, though.)

strugee:
> as updating setuptools and wheel with global pip shouldn't be an issue

I said dangerous because I think this will overwrite system package manager files which you really shouldn't do. (Though actually, Pip might bail out and require you to use some new "be unsafe, I really know what I'm doing" flag which I can't find the name of at the moment.)

strugee:
(Not that it really matters, thanks for the fix!)

PR #1763, opened 2024-05-02
Fixup install docs regarding venv

oskar456:
This fixes some tiny issues related to installation to venv

 - package `python3-venv` is necessary on Debian-based distros
 - system command `python` usually does not exist on current distros
 - pip switch `--user` cannot be used with venvs
 - fixup ExecPath in the unitfile

## How has this been tested?

Tested by doing a fresh installation on Ubuntu 24.05.

## Checklist


rixx (approved):
Thanks! We switched from `pip install --user` to venvs due to upstream changes a while back and must have missed those. (We should probably remove the package list entirely though, as I am not using any Debian-based distros, so it's always at risk of being out of date …)

Issue #1494, opened 2023-05-25
Exclamation mark notice from parallel Talks faulty


navlisilvan:
### Problem and impact

If two talks of one speaker are planned in parallel in two rooms in the programme editor, an exclamation mark indicates that this does not work. (on both talks)
In our case, however, it only shows this mark in the just moved one if the two talks (of the same length) are slightly offset.
If the talks are scheduled at exactly the same time, there will no longer be an exclamation mark in the one that has just been moved.

The exclamation mark appears and disappears in the first, not moved, talk only with reloading the page.

### Expected behaviour

An exclamation mark note for two workshops if they are held by the same speaker and overlap partially or completely.
Remove all notes if they are moved away from each other.

### Steps to reproduce

1. Create two talks from one speaker
2. Create two rooms in the programme editor
3. Schedule the two talks in the programme editor, once not overlapping, then slightly overlapping and finally completely overlapping. Then move them apart again.
4. Have a look at the exclamation mark noten

### Screenshots

![Pretalx issue](https://github.com/pretalx/pretalx/assets/17738471/9cca7790-f63a-4b0c-be47-a52442097aa4)

### Browser (software, desktop or mobile?) and version

desktop, Chrome and Firefox

### Operating system, dependency versions

windows

### Version

current github version

rixx:
Thank you for this extensive bug report! There's a complete rewrite of the schedule editor in the works, so I'll assume that this *might* get resolved with that version. I'm linking this issue in to the PR though, to make sure to double-check before we merge. #1340

navlisilvan:
Hi,
the bug is still there, the behaviour is still exactly the same as with the old scheduler. :/

PR #1767, opened 2024-05-07
overlap check: don't match with talk itself

alangecker:
after f222a57 for fixing #1494 every talk got a warning that it would overlap, but it was just "overlapping" with itself

## How has this been tested?
manual testing in the UI

## Checklist


rixx (commented):
Ah sorry, I didn't see this in time and just pushed the same change in 5a5d16c1c

PR #1770, opened 2024-05-17
Show "Full name" label on account creation field

paulmelis:
We have noticed several speakers creating accounts with only their first name, instead of their full name. This might due to being a non-native English speaker. Clearify the intention of the field. Also as apparently not everybody realizes that this name is how they get listed in the public speaker and session list.

## How has this been tested?

No, I don't have a development setup available for Pretalx

## Checklist

### DIFF
--- a/src/pretalx/person/forms.py
+++ b/src/pretalx/person/forms.py
@@ -42,7 +42,7 @@ class UserForm(CfPFormMixin, forms.Form):
         widget=forms.PasswordInput(attrs={"autocomplete": "current-password"}),
     )
     register_name = forms.CharField(
-        label=_("Name"),
+        label=_("Full name"),
         required=False,
         widget=forms.TextInput(attrs={"autocomplete": "name"}),
     )


rixx:
> Also as apparently not everybody realizes that this name is how they get listed in the public speaker and session list.

Regarding this: the profile form reads as follows. If this doesn't clarify matters to people, I doubt that a changed label will help. A certain percentage of people will just not read labels and messages, and we can't change that by writing more.

2024-05-20T20:01:37+02:00

paulmelis:
I was actually looking at the account creation page:

image

It's not uncommon when registering on a new server/service to pick a username different from your real name. I think that's what happened in our case. And then folks probably don't look in detail at the name in their profile page anymore, as they have already filled it in.

> As organisers have the power to change the appearance of the profile form/name field, I think we should think more carefully about how to do name handling, and in the meantime organisers can adjust the field's label (and help text) to suit their event.

How exactly would one change the description of the name field?

rixx:
> How exactly would one change the description of the name field?

As with all fields, you can change the label and description in the CfP editor.

> It's not uncommon when registering on a new server/service to pick a username different from your real name. I think that's what happened in our case.

Hm. I can see that, but at the same time, adding a new string for translation is also something I feel we should be cautious about.

rixx (commented):
After thinking about this a bit more, I decided against the "Full name" label, as not all events require full names – particularly in the art, hacker and online subcultures, speaker names are often nicknames/artist names etc.

Instead, I modified the registration name field to read "Name (display name)", in the hopes of clearing things up for users going using the registration form.

Issue #1761, opened 2024-05-02
Update site title in proposal wizard to include current proposal name


rixx:
This would make it much easier to work on multiple proposals in different tabs at the same time.

We could even update the title to include a "xx time left" message during the last day of submissions.

SkylerBurger:
Hello! I'd be happy to try adding this feature. I was able to set up the dev environment and poke around in the app, the docs are awesome by the way! I tried locating the proposal wizard mentioned in this issue's title but I don't think I found the right spot.

![Screenshot 2024-05-08 140855](https://github.com/pretalx/pretalx/assets/43080273/1c0df218-cb0f-4af9-811c-0bf648bbaeff)

In the screencap above I am logged in as the superuser and viewing a proposal. It appears the title is already updated in the tab.

![Screenshot 2024-05-08 141431](https://github.com/pretalx/pretalx/assets/43080273/648eb33f-01cd-4a00-a0a3-5a8561b952e2)

In this a screencap I'm logged in as a speaker with a proposal that has been associated to me. When viewing it, it appears the title of the site is already updated as shown in the browser tab. 

Apologies if I am looking in the completely wrong location. If this isn't the spot referred to in the issue would it be possible to include some more information regarding the page the feature should be implemented? Thank you!

rixx:
Thanks for taking a look at this! The "proposal wizard" is the multi step form that speakers have to go to to submit their proposal in the first place – sorry, that's a common internal naming in django-formtools for multistep forms, but I should have been clearer about this!

SkylerBurger:
I've added a PR for this issue and I'm happy to discuss any changes in scope or approach. I started with a simple JS approach before realizing this data might also be stored somewhere in a Django user session. I'm not super familiar with the user sessions but would be happy to look into it more if we want the proposal title to stay consistent in that browser tab throughout all of the remaining proposal wizard steps.

Wassaf001:
What is the proposal wizard?

PR #1777, opened 2024-05-22
[Issue 1761] Add Proposal Title to Tab of Proposal Wizard

SkylerBurger:
## Summary

This change was requested in #1761. This allows the proposal title entered into the proposal wizard to become the title of the page. That title is then populated in the browser tab to make identification of multiple proposal submissions easier across the browser.

## Changes

- Adds the `proposalTabTitles.js` script to the `submission_step.html` template.
  - The script takes the server rendered title and breaks it into its component pieces in order to save the translated "create proposal" prompt in the event the user deletes their title from the wizard and the default title needs to be restored.
  - When navigating to the proposal wizard it will quickly check for a proposal title that was placed into the wizard by the user session (in the case a user advances to the "Account" step, then navigates back to the "General" step).

## Feature Flow

1. User enters the proposal title into the proposal wizard and it becomes the title of the page and is populated in the browser tab.
Screenshot 2024-05-22 064351

2. If the user erases the proposal title from the proposal wizard, returning the original translated title prompt to the page and browser tab.
Screenshot 2024-05-22 064409

3. User advances to the "Account" step, from then on the proposal title is no longer in the browser tab. (Due to limitation of the current approach of this PR)
Screenshot 2024-05-22 064448

4. User navigates back to the "General" step where the title of the page again matches the proposal title.
Screenshot 2024-05-22 064351

### Thoughts

With the solution in this PR, the proposal title is only visible in the browser tab while the user is working on the "General" step of their proposal. I'm simply pulling the value from the input element, which is not present during the other steps. I'm fairly novice with Django but from what I was reading it looks like the form data is saved in the user's session, so if we want to have the title present across all of the proposal wizard steps we might want to look into pulling that info from the session. I figured I'd cut this PR just in case I'm moving out of scope, I'd be happy to look into getting the proposal title across all steps of the wizard if desired.

## How has this been tested?

This change was manually tested in the UI. I didn't happen to see frontend tests in the repo, if I missed them I'm happy to write tests for this feature.

## Checklist

- [ ] I have added tests to cover my changes. (Happy to add if there are frontend tests)

### DIFF
--- a/src/pretalx/static/cfp/js/proposalTabTitles.js
+++ b/src/pretalx/static/cfp/js/proposalTabTitles.js
@@ -4,7 +4,7 @@ function updateTitle(newTitle) {
     if (newTitle === "") {
         document.title = titleParts.join("::");
     } else {
-        document.title = `${newTitle}::${titleParts[1]}::${titleParts[2]}`;
+        document.title = `${newTitle} ::${titleParts[1]}::${titleParts[2]}`;
     }
 }
 
--- a/src/pretalx/static/cfp/js/proposalTabTitles.js
+++ b/src/pretalx/static/cfp/js/proposalTabTitles.js
@@ -2,9 +2,9 @@ const titleParts = document.title.split("::");
 
 function updateTitle(newTitle) {
     if (newTitle === "") {
-        document.title = titleParts.join(" :: ");
+        document.title = titleParts.join("::");
     } else {
-        document.title = `${newTitle} :: ${titleParts[1]} :: ${titleParts[2]}`;
+        document.title = `${newTitle}::${titleParts[1]}::${titleParts[2]}`;
     }
 }
 
--- a/src/pretalx/static/cfp/js/proposalTabTitles.js
+++ b/src/pretalx/static/cfp/js/proposalTabTitles.js
@@ -1,0 +1,15 @@
+// Parse and store title from the server's render since it may have been translated.
+let defaultTitleElements = document.title.split('::');
+
+if (defaultTitleElements.length !== 3) {
+  console.error('Could not parse site title while adding proposal title change listener.');
+} else {
+  document.getElementById('id_title').addEventListener('change', (event) => {
+    let newTitle = event.target.value;
+    if (newTitle === '') {
+      document.title = defaultTitleElements.join(' :: ');
+    } else {
+      document.title = `${newTitle} :: ${defaultTitleElements[1]} :: ${defaultTitleElements[2]}`;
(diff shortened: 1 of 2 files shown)


src/pretalx/static/cfp/js/proposalTabTitles.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+})
+
+if (titleParts.length !== 3) {
+    console.error(
+        "Could not parse site title while adding proposal title change listener.",
+    )
+} else {
+    document.getElementById("id_title").addEventListener("change", (event) => {
+        const newTitle = event.target.value
```
rixx, line 16: This part could be a function, like `updateTitle`, in order to avoid the repetition with `DOMContentLoaded` handler above.

src/pretalx/static/cfp/js/proposalTabTitles.js
```diff
@@ -0,0 +1,23 @@
+const titleParts = document.title.split("::")
+
+document.addEventListener("DOMContentLoaded", () => {
```
rixx, line 3: This part should probably go inside the `titleParts.length !== 3` safeguard too, just so that we don't touch the title at all if we don't know what's going on.

rixx (changes requested):
This is looking pretty good! I added a few notes, and will look into handling the Python part for the subsequent proposal wizard pages.

rixx:
Added the Python part in 2327e6a34.

src/pretalx/static/cfp/js/proposalTabTitles.js
```diff
@@ -0,0 +1,30 @@
+const titleParts = document.title.split("::");
+
+function updateTitle(newTitle) {
+    if (newTitle === "") {
+        document.title = titleParts.join(" :: ");
```
rixx, line 5: You're splitting the title parts on `::`, but you're joining on ` :: ` – my naive reading would be that every title will contain double spaces once `updateTitle` is called. Probably not very visible, but with tab title space being at a premium … could you join on `::` here and remove the spaces next to `titleParts[1]` and `2` below?

rixx (changes requested):
Last minor quibble, but looking really clean now!

Just a heads-up: I'm at an event (or worse: at the build-up phase of an event) starting tomorrow, so I'll probably take a while to get around to merging this.

SkylerBurger:
Sounds good, have a nice conference!

rixx (approved):
Thank you! :tada:

PR #1787, opened 2024-06-12
Remove pretalx.common.models.Choices

### LINKED ISSUE #1758: Remove pretalx.common.models.Choices
Our helper class for enums/choices is used in only two places, and is from a time before Django had `TextChoices`, which we should use instead.

mridubhatnagar:
Use django's TextChoice field instead of Choices class

closes #1758

Instead of creating a user defined class as Choices. We can use Django's built-in field TextChoices.

#1758 

## How has this been tested?

Tested manually

## Checklist


rixx (changes requested):
There are a number of issues with this PR:

- This PR considers only this one file, not the use in the `submission` app.
- Localisation is removed entirely.
- Have you confirmed that this code still generally works as intended?
- Contrary to the PR title, you haven't actually removed the code for the `Choices` class
- Please run our set of linters and make sure your code complies – for example, you haven't removed the unused import of the `Choices` class here

mridubhatnagar:
@rixx the changes that I had done in question.py those I had manually tested on my local. And, they were working fine. 
I have run linters as mentioned in the documentation.

Also, I didn't quite understand what is meant by localisation is completely removed.

rixx:
> I am not quite sure about the changes that would be needed in Submissions.py.

The functionality that's currently part of the `SubmissionStates` class would have to be refactored.

> Also, I didn't quite understand what is meant by localisation is completely removed.

You removed all the translation / gettext calls, so that the choice options will now not show up as translated strings when running pretalx in another language.

mridubhatnagar:
> > I am not quite sure about the changes that would be needed in Submissions.py.
> 
> The functionality that's currently part of the `SubmissionStates` class would have to be refactored.
> 
> > Also, I didn't quite understand what is meant by localisation is completely removed.
> 
> You removed all the translation / gettext calls, so that the choice options will now not show up as translated strings when running pretalx in another language.

Choices class had 2 class methods. `get_max_length` and `get_choices`. get_choices was returning list of choices. I replaced it with <classname>.choices. And, replaced `get_max_length` with the function definition of get_max_length. 

I didn't understand. How this change resulted in removal of all the translation/gettext?

mridubhatnagar:
@rixx thanks for your inputs. Based on your inputs have updated the changes that I had done in `questions.py`. 
However, for `submissions.py`. Inheriting the class from TextChoices instead of Choices is leading to changes at other places in the codebase. Hence, I am still figuring it out.

As, the class is inheriting from TextChoices. I am not able to access the dictionaries present in SubmissionStates class from outside.

PR #1789, opened 2024-06-13
Add missing migration dependency on submission.0061_cfp_settings for event.0034_fix_language_codes

d235j:
This PR fixes migration from relatively old pretalx versions, which fails due to a missing dependency (or alternatively, an incorrect check for existence of a class attribute).

## How has this been tested?

This was tested by migrating an old Pretalx instance to a newer one. The migration would fail with the following error:
```
  File "pretalx/event/migrations/0034_fix_language_codes.py", line 98, in fix_language_codes
    if not cfp.settings or "flow" not in cfp.settings or not cfp.settings["flow"]:
AttributeError: 'CfP' object has no attribute 'settings'
```

Code review / inspection also reveals the issue.

This could alternatively fixed by using code such as
```
    if not hasattr(cfp, "settings") or "flow" not in cfp.settings or not cfp.settings["flow"]:
```
in `pretalx/event/migrations/0034_fix_language_codes.py`.

## Checklist

Tests and documentation changes appear to not be applicable to this bug fix. I'm not certain of the preferred format of `doc/changelog.rst`.

### DIFF
--- a/src/pretalx/event/migrations/0034_fix_language_codes.py
+++ b/src/pretalx/event/migrations/0034_fix_language_codes.py
@@ -129,6 +129,7 @@ class Migration(migrations.Migration):
         ("person", "0014_speakerinformation"),
         ("schedule", "0013_auto_20191107_1748"),
         ("submission", "0052_auto_20201010_1307"),
+        ("submission", "0061_cfp_settings"),
         ("mail", "0009_queuedmail_locale"),
     ]
 


rixx (approved):
Thank you!

rixx:
I should note, just for completeness's sake, that this should also be fixable by updating version-by-version.

PR #1792, opened 2024-06-14
Add clarification regarding mail config

d235j:
This adds a clarification to the documentation to reduce user confusion in cases where a per-event mail configuration is already in place and the user is not aware.

## How has this been tested?

This is a documentation change. Rendering appears correct, but feel free to edit as needed.
## Checklist

Feel free to adjust wording as needed, but I believe such a message should be somewhere in the docs.

### DIFF
--- a/doc/administrator/configure.rst
+++ b/doc/administrator/configure.rst
@@ -185,6 +185,7 @@ The database section
 
 The mail section
 ----------------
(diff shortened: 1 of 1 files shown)


rixx (commented):
Thank you! I completely agree that a note here would be a good idea – however, I started writing out a long feedback note on how I'd prefer the note to be structured (both in content and in things like linewraps).

But it would be a bit ridiculous (and, for you, probably pretty annoying!) to have you rework three sentences (possibly repeatedly), so I decided to just write them myself (a3a7c77da). But please don't take that as negative feedback – I'm very thankful that you pointed out where the docs are lacking and what information you consider important!

d235j:
Thanks, that's fine with me!

PR #1802, opened 2024-07-19
Removes text area max height

ekohilas:
When overviewing submissions in edit, the max height imposed on the text creates an annoying user experience, requiring to view the submission through a small window that needs to be scrolled.

This solves the problem by removing that restriction.

## How has this been tested?

This was tested manually through editing CSS in browser.

image

## Checklist

### DIFF
--- a/src/pretalx/static/common/scss/_forms.scss
+++ b/src/pretalx/static/common/scss/_forms.scss
@@ -265,9 +265,7 @@ table .action-column {
 }
 
 .markdown-wrapper {
-  max-height: 160px;
   textarea {
-    max-height: 160px;
     border-top: none;
     border-top-left-radius: 0;
     border-top-right-radius: 0;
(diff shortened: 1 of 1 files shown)


src/pretalx/static/common/scss/_forms.scss
```diff
@@ -276,7 +274,6 @@ table .action-column {
     border: 1px solid #ced4da;
     border-top: none;
     min-height: 160px;
-    max-height: 160px;
```
ekohilas, line 279: Note that with this line change, the text area on preview is expanded to the size of the containing markdown, which may not be preferred as it's not a backward compatible change.

The other two above are fine though.

rixx (commented):
Sorry for not merging this earlier, I was doing a major css rewrite. I functionally adopted your changes now, and let's just hope that Firefox gets off their butt soon and implements auto-growing textareas soon :eyes:

PR #1805, opened 2024-07-25
Update installation.rst

LanusseMorais:
Add command to change permissions /var/pretalx to enable reverse proxy read directory.

When a reverse proxy is used, such as nginx, it must be allowed to read the static files and return them in the requests. I have validated this configuration in two productive deployments.

- [X ] I have updated the documentation.


rixx (changes requested):
This is a detail I'd like to leave up to the administrator. Giving sweeping access permissions like this should not be required, at least not on all systems – for example, you could set ownership to a hypothetical `http` group in order to permit access to the HTTP server process. `775` is a pretty broad permission that I would not like to officially recommend, sorry.

PR #1810, opened 2024-08-05
django settings variable can't start with _

dest81:
It's not possible to get `settings._VITE_IGNORE` variable.
It raises error:
```
    102     self._setup(name)
    103     _wrapped = self._wrapped
--> 104 val = getattr(_wrapped, name)
....

AttributeError: 'Settings' object has no attribute '_VITE_IGNORE'
```

Can be reproduced by:
```
$ python manage.py shell
$ from django.conf import settings
$ settings._VITE_IGNORE
```
This happens due to fact that django [uses dir()](https://github.com/django/django/blob/main/django/conf/__init__.py#L176) function to merge project_settings with global_settings. And dir() does not return non public attributes.

## Checklist

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -3,6 +3,7 @@
 Release Notes
 =============
 
+- :bug:`common` _VITE_IGNORE -> VITE_IGNORE: django settings variable can't start with _.
 - :feature:`api` The submission API now has a filter for the ``is_featured`` field.
 - :feature:`cfp,1761` In the CfP submission multi-step form, the tab title now reflects the proposal title, to make it easier to work on multiple proposal submissions at the same time.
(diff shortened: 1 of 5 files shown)


doc/changelog.rst
```diff
@@ -3,6 +3,7 @@
 Release Notes
 =============
 
+- :bug:`common` _VITE_IGNORE -> VITE_IGNORE: django settings variable can't start with _.
```
rixx, line 6: Just a heads-up: I'll remove this line from the release notes, as I think that it will only confuse the readers needlessly: the reader of the release notes is assumed to be an interested pretalx user or administrator of a self-hosted instance, and this note would serve to confuse them, without having any impact on them.

rixx (approved):
Thank you!

PR #1837, opened 2024-09-30
Add quotes around the schedule ETag value

mweinelt:
Fixes the value of the ETag field, that is currently being stripped by nginx, because it considers its value invalid per RFC2616¹, as it is missing the required quotation marks.

See also https://forum.nginx.org/read.php?2,286645,286647.

[1] https://datatracker.ietf.org/doc/html/rfc2616#section-14.19

cc @johnjohndoe

## How has this been tested?

Tested in production on talks.mrmcd.net.

```
# curl --unix-socket /run/pretalx/pretalx.sock http://talks.mrmcd.net/2024/schedule/export/schedule.xml -I
HTTP/1.1 200 OK
Server: gunicorn
[...]
ETag: "08ed860b1ef6cd0578399766e8794c8fbaf02c40"

$ curl https://talks.mrmcd.net/2024/schedule/export/schedule.xml -I
HTTP/2 200 
server: nginx
[...]
etag: "08ed860b1ef6cd0578399766e8794c8fbaf02c40"

$ curl https://talks.mrmcd.net/2024/schedule/export/schedule.xml -I --compressed
HTTP/2 200 
server: nginx
[...]
etag: W/"08ed860b1ef6cd0578399766e8794c8fbaf02c40"
```

## Checklist

### DIFF
--- a/src/pretalx/agenda/views/schedule.py
+++ b/src/pretalx/agenda/views/schedule.py
@@ -122,7 +122,7 @@ class ExporterView(EventPermissionRequired, ScheduleMixin, TemplateView):
             raise Http404()
         if request.headers.get("If-None-Match") == etag:
             return HttpResponseNotModified()
-        headers = {"ETag": etag}
+        headers = {"ETag": f'"{etag}"'}
         if file_type not in ("application/json", "text/xml"):
             headers["Content-Disposition"] = (
                 f'attachment; filename="{safe_filename(file_name)}"'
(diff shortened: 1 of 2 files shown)


mweinelt:
The django test client seems to want the bare etag value :thinking:

rixx (commented):
This probably means that we're also parsing the E-Tag incorrectly, so our caching will break on all web servers (rather just on nginx) when merging this, right?

mweinelt:
Yes, that follow-up commit looks plausible!

Issue #1596, opened 2023-10-03
Plugin integration in email sending


rixx:
Related to #1594, but going potentially deeper: It would be nice to have plugins handle emails entirely / optionally suppress the sending of emails. A use case for this would be an event that handles all speaker communications through a ticket system (like Jira, Zammad, RT, …).

Some thoughts on implementation and problems:

### Super low level: email backend

Plugins can already implement email backends, which is what is getting called by `mail_send_task`. However, email backends get basically zero information: Just the message to be sent, nothing else. So no information of the message type, or even the event – and any message from password reset email to talk acceptance runs through there. We could scope the email backend to an event with a bit of effort, but that would still leave the problem of no context.

### Mid level: `queuedmail_pre_sent` hook

Following #1594, we could implement a `queuedmail_pre_sent` hook, and give it a way to abort the actual sending of the message (probably by raising an exception), while still continuing to run the rest of the `send()` method (like setting the sent timestamp and saving the object). I'd still want to limit this hook to only handling `QueuedMail` objects with an `id`, thus allowing us to ignore any emails that are not visible to organisers (like password resets, speakers inviting other speakers, and crucially reminders to people with draft proposals, who **never** should be visible to organisers).

`QueuedMail.send()` does lack a bit of context (like the proposal an email is sent for, for example, or more generally the message context), and storing that context would be very annoying, as it basically requires generic foreign keys to a degree, but I do think it would be good enough to just … add the proposal(s) in question to a `QueuedMail` and have that be good enough. We already store the user account and the template (if any), of course.

### High level: message sending plugins

An even better abstraction, imo, would be to allow plugins to register message sending capabilities, and turning the standard email sending into one such plugin (active by default). Organisers could then even activate multiple sending plugins, configure a priority, and if messages are allowed to be sent via multiple channels. Since `QueuedMail` objects can already be displayed by ID to the user, this would make it possible to e.g. send an SMS (I know, I know) just saying "new message from the $event organisers at $url", while *also* sending an email etc.

We'd definitely need to add the same level of context mentioned in the mid-level solution, and additionally flags for quiet / non-organiser-visible sending (which would be fine in some backends, like SMS, but not in others, like Jira). We'd have to hammer out an API that is somewhat future-proof, but I imagine that would largely be a mix of the `to_mail()` and the `send()` signatures.

[...]

PR #1869, opened 2024-10-27
Add custom headers to outgoing mails to track event slug and submission code (if available).

badbadc0ffee:
This makes it possible to link back into Pretalx from ticket management systems like RT if they are used as replacement for the normal SMTP backend. A plugin 'pretalx-rt' is currently developed therefore in preparation of the 38c3.

The change has been tested with the pretalx-rt plugin which is under development.
Without that plugin activated the change remains mostly invisible to users.
Email sent by Pretalx will contain two extra headers which could be useful also for other integrations (and for testing).

The extra header 'X-Pretalx-Event' will always be present if an email is related to an event in Pretalx. It will contain the events slug. Only emails that would be sent out and that are not related to events (i.e. update notifications)  would lack this header.
The extra header 'X-Pretalx-Code' will be set to the submission code if an email is related to a submission.

## Checklist


badbadc0ffee:
Does this need additional documentation?

rixx (changes requested):
I’m not convinced that adding this feature to pretalx core is a good idea – we do have #1596 open as a discussion on how to give plugins more input on email handling, and I think that would be a better solution all around, as these headers are pretty specific to your plugin.

At the same time, *if* we end up adding these headers via pretalx, there are a couple of open questions that would need to be covered first: How do you want to deal with emails that concern multiple proposals – add the same header twice? And, both proposals and speakers have a `code` field used as ID, so seeing ´X-Pretalx-Code` doesn't make the object type referenced clear.

Also (again, if we end up putting this in pretalx-core), I don’t think that putting these headers into a JSON field is the best fit for the data, and instead the headers should be set at the time of sending the email. The event information is already available dynamically via the `QueuedMail.event` field – and I think we would be better served by adding a `QueuedMail.submissions` relation, which would also be generally useful in linking emails with proposals, not just with speakers.

But like I said, I’m not convinced that adding these headers in pretalx-core is the best direction in the first place, and I think giving plugins better input on sent emails is the better idea here – possibly combined with a `QueuedMail.submissions` field in pretalx-core in order to make the required information available to the plugin.

rixx:
Before I forget the next time I look at this, I should also add that the `X-` convention for custom headers is discouraged and instead:


> 1.  SHOULD assume that all parameters they create might become
       standardized, public, commonly deployed, or usable across
       multiple implementations.
>   2.  SHOULD employ meaningful parameter names that they have reason to
       believe are currently unused.
>   3.  SHOULD NOT prefix their parameter names with "X-" or similar
       constructs.
>
>   Note: If the relevant parameter name space has conventions about
   associating parameter names with those who create them, a parameter
   name could incorporate the organization's name or primary domain name
   (see [Appendix B](https://www.rfc-editor.org/rfc/rfc6648#appendix-B) for examples).

badbadc0ffee:
> 3. SHOULD NOT prefix their parameter names with "X-" or similar constructs.
That is really easy to fix. ;-)
I actually wasn't aware of RFC 6648, I'm so old, I am more of a RFC 822 person...

badbadc0ffee:
> we do have #1596 open

I also wasn't aware of the very good summary of options in that ticket.
This approach is obviously addressing the 'super low level' approach.
How about making the extra headers
- compliant to RFC 6648
- use clear names for submissions and
- add references to persons (also using their 'code')
- support multiple references to submissions and/or persons in one message
- configurable and inactive by default
 
That would allow us to quickly create a working plugin to interact with RT and at the same time, I'd propose to also work on the 'mid level approach' and upgrade the plugin once that approach (definitely more elegant) is ready.

badbadc0ffee:
obsolete, see #1871 instead

PR #1871, opened 2024-10-28
Add submissions to QueuedMail

badbadc0ffee:
This shall make it possible to link back into Pretalx from ticket management systems like RT once messages get sent out to users via such systems.

This is one part of the changeset in PR #1869.

This part only addresses the addition of submissions to the QueuedMails. This can be either used for a 'super low level approach' (i.e. making extra headers and replacing the mail backend) or also (preferably) higher level integration approached as discussed in PR #1596.

## Checklist


badbadc0ffee:
This PR also contains a "mid level queuedmail_pre_sent hook" as mentioned in #1596 now, the extended email headers as suggested in #1869 are thus obsolete.

src/pretalx/mail/signals.py
```diff
@@ -8,6 +8,18 @@
 As with all event-plugin signals, the ``sender`` keyword argument will contain the event.
 """
 
+queuedmail_pre_send = EventPluginSignal()
+"""
+This signal is sent out before a ``QueuedMail`` will been sent.
```
rixx, line 13: It might be good to explain here that this signal will only be sent for emails that will be saved, so it will not be triggered for emails that are fire-and-forget for privacy reasons, like password resets or reminders about draft proposals.

src/pretalx/mail/signals.py
```diff
@@ -8,6 +8,18 @@
...
+Receivers may set the ``sent`` timestamp to skip sending via the regular
+email backend but shall not alter any other data of the ``QueuedMail``
+instance.
```
rixx, line 16: We can't really enforce the altering of other data here, so we may as well permit it, I think. Adding something like a custom signature to the email text, or adding/changing attachments, is something I could imagine as a use case.

src/pretalx/mail/models.py
```diff
@@ -309,22 +316,27 @@ def send(self, requestor=None, orga: bool = True):
...
-                "event": self.event.pk if has_event else None,
-                "cc": (self.cc or "").split(","),
-                "bcc": (self.bcc or "").split(","),
-                "attachments": self.attachments,
-            },
-            ignore_result=True,
-        )
-
-        self.sent = now()
+            queuedmail_pre_send.send(
```
rixx, line 319: Let’s move the signal call above `text = self.make_text()`, so that the signal consumers can actually change the content of the email being sent, I think that would be nice.

src/pretalx/mail/models.py
```diff
@@ -162,6 +162,9 @@ def to_mail(
             )
             if commit:
                 mail.save()
+                if "submission" in context_kwargs:
+                    submission = context_kwargs["submission"]
+                    mail.submissions.add(submission)
```
rixx, line 167: We could additionally also add an optional `submissions=None` kwarg to this method, and use that to add passed submissions additionally.

Not sure this is a good idea – `to_mail` already has a long signature, and making it longer isn’t great. But I don’t really see a better way of actually adding multiple submissions here, which would be useful e.g. from the `Schedule.generate_notifications` method, which is (I think) the currently only place where we explicitly send emails concerning multiple proposals, and it would be nice to link those emails to the proposals included. What do you think?

rixx (commented):
Thanks, this looks really good! Give me another couple of days to think about the signal and if I can see any potential problems with it (re leaking data / calling the signal when we shouldn’t).

I left some comments – one is a question, and the others are optional, if you don’t get around to them, I can handle them after the merge, too.

rixx:
Just a note so I don’t forget: After this has been merged, I want to show the linked proposals in the outbox/sent mail tables, colour-coded with their track colour, to make it easier for organisers that feel mostly responsible for just one track to see which outgoing emails are relevant to them.

src/pretalx/mail/signals.py
```diff
@@ -8,6 +8,18 @@
...
+
+As with all event-plugin signals, the ``sender`` keyword argument will
+contain the event. Additionally, the ``mail`` keyword argument contains
+the ``QueuedMail`` instance itself.
```
rixx, line 20: One more thought: We should figure out how to deal with exceptions raised by receivers of this signal.

Two real options:

- ignore exceptions – use the `send_robust` sending method, and document that we will disregard exceptions
- allow exceptions to abort the sending of emails: this is nicer for plugin authors, but at the same time, we don’t have a channel back to the user in most cases, as emails are sent asynchronously.

Since a plugin blocking the sending of an email has no way to tell the user **that** and **why** it has blocked the sending, I’m leaning towards the first option.

badbadc0ffee: I’m also leaning towards the first option. I thought about this a little bit already and didn't really have a better idea.

badbadc0ffee:
> I want to show the linked proposals in the outbox/sent mail tables

Nice. I've already prepared a proposal to add optional "badges" in these tables for references to RT, btw.
I'd still propose to make that optional: some users just recently mentioned that the "featured" column int the submissions table is too much for them as we don't use the feature.

rixx:
> I'd still propose to make that optional: some users just recently mentioned that the "featured" column int the submissions table is too much for them as we don't use the feature.

I can see that, yeah – I am still considering just switching all pretalx tables/list views to django-table2, which would give us user-selectable columns for free (this is how NetBox does it). I even have an [old PR](https://github.com/pretalx/pretalx/pull/1589) for it, but it added 150-400+ms in page load and I still can’t quite myself take that performance hit.

Possibly time to revisit that, though, because it *would* give us trivial overall sorting and persisted user-selected columns, which seems really neat …

rixx (commented):
GitHub is being annoying about rebases and conflicts, so this is superseded by #1889

Issue #1879, opened 2024-11-01
Badges (i.e. in the email outbox list) lack their background color.


badbadc0ffee:
### Problem and impact

Badges are used to highlight mail generated from certain templates.
The css is missing definitions for badge-succee, badge-warning, etc.
(This is probably due to the removal of the bootstrap sccs files.)
Badges are invisible now unless you know what to look for an hover the area...

### Expected behaviour

Badges show up with the respective background colors.

### Steps to reproduce

1. Create an acceptance email.
2. Check the email outbox.
3. Try to find the 'Accept email' badge.

### Link

/orga/event/{slug}/mails/outbox/

### Version

2024.3.0 ff.

PR #1880, opened 2024-11-01
Add background color to badges

badbadc0ffee:
This PR resolves issue #1879. It does so by adding missing css definitions to `base.css`

## How has this been tested?
Changed locally, checked in web browser.

### DIFF
--- a/src/pretalx/static/common/css/base.css
+++ b/src/pretalx/static/common/css/base.css
@@ -316,35 +316,41 @@ small {
 /* Conventions and custom elements */
 
 .color-success,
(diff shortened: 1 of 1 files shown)


rixx (commented):
hmm, I think I want to actually get rid of the extra classes that are kind of useless – I’ll switch the few remaining occurrences to using `color-` instead of `badge-`. Thank you for pointing this out, though!

badbadc0ffee:
Sure sounds good. In case you'd like to cherry-pick, here is an updated one...

PR #1885, opened 2024-11-05
Add html and forms signals for plugin integration

badbadc0ffee:
This is an initial draft / a proposal:
I'd like to add some signals to be able to integrate information related to emails and submissions from a plugin.

I'd no longer call this a draft now but rather a proposal. ;-)

- The number of html_signals has now been reduced to one per domain (speaker, submission, mail). - Well actually there is one additional hook for a link where I'd love to see what people like to use more during the 38c3 before I decide if I'd better put a link to external systems within the form data or in the corner where other links live.
- The mail badges have been removed because of the performance impact of that approach.
- Additionally, form signals have been included for the same domain to allow extending the form including the possibility to edit data provided by plugins.

Is this going into the right direction? I'd really love to merge the changes in the plugins for RT, Frab, Zammad and Rocket which are in preparation as well...

### DIFF
--- a/src/pretalx/mail/signals.py
+++ b/src/pretalx/mail/signals.py
@@ -37,3 +37,14 @@ As with all event-plugin signals, the ``sender`` keyword argument will
 contain the event. Additionally, the ``mail`` keyword argument contains
 the ``QueuedMail`` instance itself.
 """
+
+mail_form_html = EventPluginSignal()
+"""
+This signal is sent out to display additional information related to emails in
+the internal organiser area.
+
+As with all plugin signals, the ``sender`` keyword argument will contain the
+event. Additionally, the signal will be called with the ``request`` it is
+processing, and the ``mail`` which is currently displayed.
+The receivers are expected to return HTML.
+"""
--- a/src/pretalx/mail/signals.py
+++ b/src/pretalx/mail/signals.py
@@ -48,3 +48,14 @@ event. Additionally, the signal will be called with the ``request`` it is
 processing, and the ``mail`` which is currently displayed.
 The receivers are expected to return HTML.
 """
+
+mail_forms = EventPluginSignal()
+"""
+This signal is sent out to inject additional form fields on the submission
+pages in the internal organiser area.
+
+As with all plugin signals, the ``sender`` keyword argument will contain the
+event. Additionally, the signal will be called with the ``request`` it is
+processing, and the ``mail`` which is currently displayed.
+The receivers are expected to return one or more forms.
+"""
--- a/src/pretalx/person/signals.py
+++ b/src/pretalx/person/signals.py
@@ -1,6 +1,17 @@
 from django.dispatch import receiver
 
-from pretalx.common.signals import register_data_exporters
+from pretalx.common.signals import EventPluginSignal, register_data_exporters
+
+speaker_form_html = EventPluginSignal()
+"""
+This signal is sent out to display additional information on the speaker
+pages in the internal organiser area.
+
+As with all plugin signals, the ``sender`` keyword argument will contain the
+event. Additionally, the signal will be called with the ``request`` it is
+processing, and the ``person`` which is currently displayed.
+The receivers are expected to return HTML.
+"""
 
 
 @receiver(register_data_exporters, dispatch_uid="exporter_builtin_csv_speaker")
--- a/src/pretalx/person/signals.py
+++ b/src/pretalx/person/signals.py
@@ -14,6 +14,18 @@ The receivers are expected to return HTML.
 """
 
 
+speaker_forms = EventPluginSignal()
+"""
+This signal is sent out to inject additional form fields on the submission
+pages in the internal organiser area.
+
+As with all plugin signals, the ``sender`` keyword argument will contain the
+event. Additionally, the signal will be called with the ``request`` it is
+processing, and the ``person`` which is currently displayed.
+The receivers are expected to return one or more forms.
+"""
+
+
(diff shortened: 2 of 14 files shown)


badbadc0ffee:
https://github.com/badbadc0ffee/pretalx-signals-demo contains a small demo that emits a sample for each of the suggested signals

src/pretalx/mail/signals.py
```diff
@@ -37,3 +37,25 @@
...
+As with all plugin signals, the ``sender`` keyword argument will contain the
+event. Additionally, the signal will be called with the ``request`` it is
+processing, and the ``mail`` which is currently displayed.
+The receivers are expected to return HTML.
+"""
+
+html_below_mail_subject = EventPluginSignal()
+"""
+This signal is sent out to display additional information related to emails in
+the internal organiser area.
```
rixx, line 55: This is not correct as implemented – the `mail_log.html` file is show to the speaker aswell.

src/pretalx/orga/templates/orga/mails/outbox_list.html
```diff
@@ -102,6 +103,7 @@ <h2>
                                 {% elif mail.template == request.event.question_template %}
                                     <span class="badge color-info">{% translate "Unanswered questions reminder" %}</span>
                                 {% endif %}
+                                {% html_signal "pretalx.mail.signals.html_after_mail_badge" sender=request.event request=request mail=mail %}
```
rixx, line 106: I worry about the performance hit for calling a signal on every row of a potentially 100+ row table, especially as plugins can't add fields to be prefetched/preselected, which sounds like a recipe for n+1 queries.

This kind of request makes me think that maybe we should move to using `django-tables2` (or a ripoff) regardless of the unfortunate performance penalty detailed [here](https://github.com/pretalx/pretalx/pull/1589), as then we could give plugins the option to register proper table columns instead.

src/pretalx/person/signals.py
```diff
@@ -1,6 +1,28 @@
...
+html_above_person_form = EventPluginSignal()
+"""
+This signal is sent out to display additional information on the speaker
+pages in the internal organiser area.
+
+As with all plugin signals, the ``sender`` keyword argument will contain the
+event. Additionally, the signal will be called with the ``request`` it is
+processing, and the ``person`` which is currently displayed.
+The receivers are expected to return HTML.
+"""
```
rixx, line 14: I’m not sure I like this – this way we’d accumulate a lot of these signals. I think it’s worth taking a look at what [pretix](https://docs.pretix.eu/en/latest/development/api/general.html#id1) is doing in instances like this – pretix allows plugins to add more forms to given pages, or to even replace the form class of a certain page, see e.g. `pretix.control.signals.item_forms`, `pretix.control.signals.voucher_form_class`.

In any case, I’d probably restrict this to a `pretalx.orga.speaker_form_html` signal that is rendered below the form, and if there is a super urgent need for the plugin to add content to the top of the page, it could still move content up via JS, I suppose.

badbadc0ffee: Thanks a lot for pointing that out! That approach is definitely more elegant. I'll dig into that.
The reason for me to put one hook above and one below here was that we actually have the usecase that we have some information that belongs below the form. Information from other systems like RT in particular but also Zammad and Frab. They are obviously all read-only now, btw. which could be adressed with the approach you suggested.
But we will also have some information that will be 'urgent' and should be on top of the page: during the congress, when a speaker has not shown up short before the presentation, that is the kind of indication I'd like to see above anything else.

src/pretalx/mail/signals.py
badbadc0ffee: That is right! - And should not necessarily be that way.
There obviously is already a differentiation, because the 'accordion' cannot be opened on the public profile page.
Makes a lot of sense, not to add the information there - although it might also make at least some sense to let the plugin decide.

badbadc0ffee: Minor additional remart: the accordion doesn't open in the public view but the HTML is the same as in the organiser backend.

badbadc0ffee: I have changed this and moved the hooks to inject HTML and/or forms into the email editor.
Tje hooks are not on the new email composition page now. I'm not yet sure if I also want them there.

src/pretalx/person/signals.py
badbadc0ffee: I still don't really like the idea to reorder the page using JS but removed the 'above_xxx' hooks.

badbadc0ffee: > ... I’d probably restrict this to a `pretalx.orga.speaker_form_html` signal ...

Personally, I'd prefer to have the signals in the namespace of the ,odels they are related to. But the couly of courde all be moved into the  `pretalx.orga` space if you like that more!

badbadc0ffee:
just split the PR back into smaller changes to make it a bit easier to digest. ;-)

badbadc0ffee:
I stumbled across #1959 and added support for plugins to hook into the review page template.

Now that I've added the additional signals, I'm wondering whether introducing separate signals for different locations is actually the best approach, or if a more generic signal for plugin extensions—and another one for HTML content—might be better. In that case, plugins would have to decide where to insert their content based on the request, similar to the older, existing `html_signal`.

@rixx, what would be your preference?

rixx:
Agreed that the newer more minimal approach is better – definitely better for pretalx, as it leads to cleaner and more minimal code; and also better for plugin authors wanting to add content to more than just one type of page, imo. Closing in favour of #2017, where I added a longer comment re: timeline.

PR #1887, opened 2024-11-08
Send submission_state_change signal on create

### LINKED ISSUE #1883: Event submission_state_change is not sent when a submission is initially placed
### Problem and impact

When a submission is initially made through the CFP wizard, no `submission_state_change` event is sent when the submission changes from its draft state to submitted for the first time.
The event should also be sent in this situation to ease the integration of issue tracking systems like Request Tracker in order to create an issue every time a new proposal is made in pretalx.

### Expected behaviour

When a person submits a proposal, the `submission_state_change` should immediately be sent once the submission state changes from `draft` to `submitted` even before the first notifications are sent out.
This is in particular necessary when an integration with an issue tracking system expects to find an existing issue related to the submission when the submission related notifications shall be stored in the context of that issue.

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

_No response_

rixx:
This PR is an alternative to #1884 and fixes #1883

### DIFF
--- a/src/pretalx/submission/models/submission.py
+++ b/src/pretalx/submission/models/submission.py
@@ -357,6 +357,17 @@ class Submission(GenerateCode, PretalxModel):
 
     update_duration.alters_data = True
 
+    def save(self, *args, **kwargs):
+        is_creating = not self.pk
+        super().save(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


badbadc0ffee:
Excewllent. Works flawlessly.

PR #1893, opened 2024-11-19
Add big button for organiser area if allowed

johanvdw:
As an organiser, there is usually little point in going to the CfP, and it it is not very clear you can go to the organiser area by clicking your username(I've had people asking it). This PR proposes to a big button, which is convenient.

## How has this been tested?
- we used this last year for our event
image

## Checklist

### DIFF
--- a/src/pretalx/cfp/templates/cfp/event/index.html
+++ b/src/pretalx/cfp/templates/cfp/event/index.html
@@ -31,6 +31,11 @@
                     {{ phrases.cfp.go_to_cfp }}
                 </a>
             {% endif %}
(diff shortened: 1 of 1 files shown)


rixx (approved):
hm, this will make organisers see up to five buttons (or realistically four buttons, admittedly) there, which may look a bit silly. Also, this adds further disparity between the very similar-seeming buttons on the event start page and the event cfp page, which are just different enough to be uncanny. But I suppose that's still better than not having the button at all.

rixx:
Thank you!

Issue #1900, opened 2024-11-27
HTML export not working if schedule is not public


badbadc0ffee:
### Problem and impact

The remark `Unset to hide your schedule, e.g. if you want to use the HTML export exclusively.` for the option `Show schedule publicly` suggest that you can have your schedule unpublished but still be able to create HTML exports.

The export unfortunately does not work that way but instead reflects exactly what an unauthenticated user sees: a bunch of 404 error messages.

### Expected behaviour

The expected behavior is exactly what the remark mentioned above also suggests: even if you decide not to publicly show your schedule within pretalx, you should still be able to export ist.

### Steps to reproduce

1. Uncheck `Show schedule publicly`
2. run `python -m pretalx export_schedule_html`
3. Check exported files → they just contain the 404 error pages.
4. Check `Show schedule publicly`
5. run `python -m pretalx export_schedule_html` abain
6. Check exported files again → they contain the correctly rendered pages.

### Version

v2024.3.1 and main

badbadc0ffee:
Still only works if triggered via commandline, not when an export is triggered via the orga web frontend.

rixx:
That sounds weird, as the web frontend also calls the manage command in the end. But I'm happy to move all functionality from the manage command to an `agenda/export.py` and call that from both the task and the manage command for a cleaner isolation and less redirection through `call_command`.

PR #1907, opened 2024-12-01
rewrite relative paths in HTML export

badbadc0ffee:
This PR addresses issue #1900. It does so by rewriting absolute paths in exported HTML files.
Additional absolute paths still exist e.g. in exported javascript and must be rewritten as well.

This is missing (at least):
- [x] move `schedule/nojs` to `schedule/nojs/index.html`
- [ ] make paths in javascript files for schedule.json relative paths
- [ ] make paths in javascript files for ForkAwesome fonts relative paths
- [x] use nojs pages by default

### DIFF
--- a/src/pretalx/agenda/management/commands/export_schedule_html.py
+++ b/src/pretalx/agenda/management/commands/export_schedule_html.py
@@ -131,7 +131,19 @@ def make_relative(base_url, content):
         )
         return relative_link
 
-    absolute_link_pattern = re.compile(rf"(?>href|src)=([\"'])(\/.*?)\1")
+    absolute_link_pattern = re.compile(r"(?>href|src)=([\"'])(\/.*?)\1")
+    return absolute_link_pattern.sub(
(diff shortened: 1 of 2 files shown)


rixx (changes requested):
Is this the best approach to this? I feel like since we already run all HTML pages that pretalx serves through a parser anyway, wouldn't it be better to use the parsed information to find URLs to replace?

In any case, we should take this action only on pretalx-generated HTML (and CSS, where needed – we already find URLs in CSS via `find_urls`. By checking the output for "DOCTYPE html", you a) potentially increase the runtime by a lot, because if that string is *not* present, the whole potentially very long string will be checked, and b) you'll match e.g. user-uploaded HTML and rewrite it, which we really should not do.

(a) could be alleviated by only checking `content[:20]` for the substring match, but b) still stands …)

badbadc0ffee:
> Is this the best approach to this?

Quite certainly not! This is a hack that tries to work on short notice.
Lets absolutely discuss and find a better solution!

Issue #1914, opened 2024-12-04
Review within one track


johanvdw:
### Problem and impact

When you are a reviewer for different tracks (in my  case main, lightning talks, bof rooms for FOSDEM), you will get a random track when you click next. 
I think in most cases when reviewing a track you prefer keep getting proposals from the same track until they run out. Eg in our case the requirements are quite different, and reviewing bof is something we will do much later.

I'm willing to implement this, but creating an issue first because someone may have conflicting use cases I'm not aware of.

Issue #784, opened 2019-09-24
Better "go to next" review workflow


rixx:
This is an issue that is well-suited for external contributors – let me know if I can be of any help, either via this issue or via support@pretalx.com.

## What this is about

pretalx allows reviewers to review submissions, which is often something reviewers do in a big session with lots of reviewing in a row. Currently, you can select either "save" or "save and next" when you're done with a review, which is not ideal.

There are a bunch of review modes that might be worth exploring and/or offering to reviewers:

- Save (and stay on page)
- Save and next (random)
- Save and next (in same track)
- Save and next (from those that have few reviews)
- Save and back to review dashboard

Offering all of these choices in the same button would declutter the interface a lot I think. Offer them as a drop-down on the button, and then remember what this user chose the last time, to keep the interface consistent.

For added convenience we should probably retain search/order/filter parameters from the review dashboard page across these calls, so that they can be reused.

## How to get started

First of all, leave a comment on this issue to make sure nobody else starts working on this issue at the same time!

Next, you need pretalx running locally on your machine. First, [fork](https://github.com/pretalx/pretalx/fork) the repository to your own GitHub account. Then clone it locally (or [set it as your remote](https://help.github.com/articles/configuring-a-remote-for-a-fork/) if you already cloned it earlier)

Follow the [developer setup guide](https://docs.pretalx.org/en/latest/developer/setup.html), and then play around with pretalx a bit.
Have a look at the places impacting this issue. Look at the code responsible for the issues (searching for relevant words or files should help you find your way around). We have an explanation of our [code structure](https://docs.pretalx.org/en/latest/developer/concepts.html) and the actual [data models](https://docs.pretalx.org/en/latest/developer/interfaces/models.html).

## How to do this

The interface to be changed is in `pretalx/orga/templates/orga/submission/review.html`, and the corresponding code in `pretalx.orga.views.review.ReviewSubmission`. 

You'll want to change the behaviour of the `get_success_url` method, and also update the method used in `request.session`, and then show the saved/selected method as the default on the next review page.

## Submit the PR

There are a number of tutorials on the Pull Request workflow online, for example this [short version](https://guides.github.com/introduction/flow/) by GitHub itself and this [more in-depth tutorial](https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github).

[...]

PR #1915, opened 2024-12-04
Stay in same track while reviewing

johanvdw:
cfr #1914

### DIFF
--- a/src/pretalx/orga/views/review.py
+++ b/src/pretalx/orga/views/review.py
@@ -648,11 +648,20 @@ class ReviewSubmission(ReviewViewMixin, PermissionRequired, CreateOrUpdateView):
             self.request.event,
             self.request.user,
             ignore=ignored_submissions,
+            track=self.submission.track
         ).first()
         if not next_submission:
             ignored_submissions = (
                 [self.submission.pk] if action == "skip_for_now" else []
             )
+            next_submission = Review.find_missing_reviews(
+                self.request.event,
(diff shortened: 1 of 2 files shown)


rixx:
Not reviewing yet due to the draft status, but I do think this is a good idea fwiw.

UI-wise, I have for a long time thought that the "next" button could use a drop-down (similar to GitHub's merge option drop-down) to change review modes (see #784). One of those could be "next in same track" – I don't think we need a track *select* so much as the option to stay within one track. What do you think?

johanvdw:
Strictly speaking, this is already ready for review. It works like this, and I also prefer this rather than the previous mode.
The reason I made it draft was because I believe other options are definitely possible.

rixx:
Another thing to think about: Do we want the progress bar to show the progress within the current filter (per-track, or later potentially per-search etc), or do we want to keep the progress bar to show all reviewable proposals?

In the same vein, we need to think about what to do when the end of the currently selected track is reached: Show the "you’re all done" message / continue with other reviews / etc. (It's always UI questions that are the hard ones …)

rixx:
Sorry for only now getting back to this – I was just reviewing all open PRs in preparation for the big v2025.1.0 release. I want to give this some more attention, which means it won’t be part of the upcoming release. Sorry about that –this is definitely on me; it’ll be part of the subsequent release (which I hope won’t take nearly as long, as I have a bunch of other small-but-nice changes queued up for it).

Issue #1910, opened 2024-12-03
JSON Exporter uses deprecated field `public_name`


tribut:
### Problem and impact

When exporting the schedule as JSON (frab compatible), the speakers name is (only) exported as `public_name`. According to the schema that is also included in this repo, this field is deprecated: https://github.com/pretalx/pretalx/blob/d7b427a3e8050efd94d513b73e69f37f1908839a/src/tests/fixtures/schedule.json#L345-L348

### Expected behaviour

Use `name` instead of `public_name`:

https://github.com/pretalx/pretalx/blob/d7b427a3e8050efd94d513b73e69f37f1908839a/src/pretalx/schedule/exporters.py#L228

rixx:
Closed in f2e147b8cb

PR #1925, opened 2024-12-18
Replace public_name with name

Bhoomish-Patel:
# Replace `public_name` with `name` in schedule export

The `public_name` field in the schedule export JSON is marked as deprecated in the schema and should be replaced with the `name` field for schema compliance. 

This PR closes issue #1910. It does so by updating the schedule export logic to use the `name` field instead of the deprecated `public_name`.

## Checklist

- [ ] My change is listed in the `doc/changelog.rst`.


rixx (commented):
I don't think it's a good idea to remove the property just because the upstream has deprecated it. Downstream users may still rely on its presence.

PR #1927, opened 2024-12-20
schedule.models.slot.TalkSlot: improve generation of frab_slug

### LINKED ISSUE #1926: use python-slugify for talkslot.frab_slot
Currently pretalx produces slugs that are not "pretty" (aka missing umlauts, dashes at the end etc). We could replace this with `python-slugify` to get prettier slugs.

I'm happy to provide a pull request for that, just wanted to ensure i don't miss a obvious caveat when using `python-slugify`

Kunsi:
This removes trailing dashes (was underscores, which were not allowed there anyway). Also adds some replacements for special characters that were removed previously.

This fixes #1926

## Checklist

### DIFF
--- a/src/pretalx/schedule/models/slot.py
+++ b/src/pretalx/schedule/models/slot.py
@@ -1,6 +1,7 @@
 import datetime as dt
 import re
 import string
+import unicodedata
 import uuid
 from urllib.parse import urlparse
 from zoneinfo import ZoneInfo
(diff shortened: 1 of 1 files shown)


src/pretalx/schedule/models/slot.py
```diff
@@ -168,11 +193,13 @@ def id_suffix(self):
...
+        title = title.lower()
+        for search, replace in SLUG_ALTERNATIVE_TRANSLATIONS.items():
+            title = title.replace(search, replace)
         legal_chars = string.ascii_letters + string.digits + "-"
         pattern = f"[^{legal_chars}]+"
         title = re.sub(pattern, "", title)
-        title = title.lower()
-        title = title.strip("_")
+        title = title.strip("-")
         return f"{self.event.slug}-{self.submission.pk}{self.id_suffix}-{title}"
```
rixx, line 204: We should strip this string of `-` too, or just return the plain string in case the title is empty.

rixx (commented):
Looks good – I wonder if it might make sense to instead do

    unicodedata.normalize("NFD", text).encode("ASCII", "ignore")

Which we use in `pretalx/common/text/path.py`. This would leave out `ß` completely, and transliterate `ä` as `a`, which may be a bit worse, but would remove the need to keep a transliteration table of our own, which I think is better.

This *would* still make all purely cyrillic titles appear as empty (so the frab slug would be still only event slug + submission pk, but I think that's just how things are going to be.

johanvdw:
>hy is the event slug actually added to the slug? It seems redundant to me.

Kunsi:
Oh, totally forgot this existed, will take a look soon.

rixx (approved):
Thank you!

PR #1931, opened 2025-01-05
Add additional signals and setting for auth plugins

adamskrz:
I'd like to add a few signals to allow the addition of extra login methods by plugins (such as [pretalx-social-auth](https://github.com/adamskrz/pretalx-social-auth)). I've created two html signals for adding to the login page, and for adding to the user profile page. Additionally, there's a signal that's called when a user's account is deactivated, allowing plugins to delete any personal information stored in their own models.

I've also added an `additional_auth_backends` setting to the config file.

Happy to receive any feedback as I'm sure these aren't necessarily the best ways to achieve this!

## Checklist

### DIFF
--- a/src/pretalx/common/templates/common/auth.html
+++ b/src/pretalx/common/templates/common/auth.html
@@ -58,5 +58,5 @@
         {% endif %}
     </div>
 {% endif %}
-{% html_signal "pretalx.common.signals.auth_html" sender=request.event %}
+{% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request %}
 {% if no_form %}</div>{% else %}</form>{% endif %}
--- a/src/pretalx/common/templates/common/auth.html
+++ b/src/pretalx/common/templates/common/auth.html
@@ -1,6 +1,7 @@
 {% load compress %}
 {% load i18n %}
 {% load static %}
+{% load html_signal %}
 
 {% include "common/forms/errors.html" %}
 {% if no_form %}
@@ -57,4 +58,5 @@
         {% endif %}
     </div>
 {% endif %}
+{% html_signal "pretalx.common.signals.auth_html" sender=request.event %}
 {% if no_form %}</div>{% else %}</form>{% endif %}
--- a/src/pretalx/common/templates/common/auth.html
+++ b/src/pretalx/common/templates/common/auth.html
@@ -58,5 +58,5 @@
         {% endif %}
     </div>
 {% endif %}
-{% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request %}
+{% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request next_url=next_url %}
 {% if no_form %}</div>{% else %}</form>{% endif %}
--- a/src/pretalx/common/templates/common/auth.html
+++ b/src/pretalx/common/templates/common/auth.html
@@ -40,6 +40,7 @@
                 {{ phrases.base.password_reset_heading }}
             </a>
         {% endif %}
+        {% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request next_url=next_url %}
     </div>
 {% endif %}
 {% if not hide_register %}
@@ -58,5 +59,4 @@
         {% endif %}
     </div>
 {% endif %}
-{% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request next_url=next_url %}
 {% if no_form %}</div>{% else %}</form>{% endif %}
--- a/src/pretalx/common/templates/common/auth.html
+++ b/src/pretalx/common/templates/common/auth.html
@@ -40,7 +40,7 @@
                 {{ phrases.base.password_reset_heading }}
             </a>
         {% endif %}
-        {% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request next_url=next_url %}
+        {% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request next_url=next_url|default:get_success_url %}
     </div>
 {% endif %}
 {% if not hide_register %}
--- a/src/pretalx/person/signals.py
+++ b/src/pretalx/person/signals.py
@@ -1,4 +1,4 @@
-from django.dispatch import receiver
+from django.dispatch import Signal, receiver
 
 from pretalx.common.signals import register_data_exporters
 
@@ -8,3 +8,13 @@ def register_speaker_csv_exporter(sender, **kwargs):
     from pretalx.person.exporters import CSVSpeakerExporter
 
     return CSVSpeakerExporter
+
+
+deactivate_user = Signal()
+"""
+This signal is sent out when a user is deactivated (i.e. deleted on the
+frontend).
+
+You will get the user as a keyword argument ``user``. Receivers are expected to
+delete any personal information they might have stored about this user.
+"""
--- a/src/pretalx/cfp/templates/cfp/event/user_profile.html
+++ b/src/pretalx/cfp/templates/cfp/event/user_profile.html
@@ -84,6 +84,8 @@
 
     {% include "common/user_api_token.html" %}
 
+    {% html_signal "pretalx.common.signals.profile_bottom_html" sender=request.event user=user %}
+
     <div>&nbsp;</div>
     <h3>{% translate "Account deletion" %}</h3>
     <form action="{{ request.event.urls.user_delete }}" method="post" class="form">
--- a/src/pretalx/common/signals.py
+++ b/src/pretalx/common/signals.py
@@ -237,7 +237,7 @@ Make sure that any user content in the HTML code you return is properly escaped!
 
 As with all event-plugin signals, the ``sender`` keyword argument will contain the event.
 """
-auth_html = EventPluginSignal()
+auth_html = django.dispatch.Signal()
 """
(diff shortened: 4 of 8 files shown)


src/pretalx/common/templates/common/auth.html
```diff
@@ -39,6 +40,7 @@ <h4 class="text-center">{% translate "I already have an account" %}</h4>
                 {{ phrases.base.password_reset_heading }}
             </a>
         {% endif %}
+        {% html_signal "pretalx.common.signals.auth_html" sender=request.event request=request next_url=next_url|default:get_success_url %}
```
adamskrz, line 43: This is the part I'm probably the least happy with - I added get_success_url to the context to work in forms, and it uses next_url in the cfp account step. It seems quite crude to find the redirect url in the template like this.

tjarbo:
This is a very important PR in my opinion, how can I help to get this merged?
SSO is all over the internet and while I usually would like to see it be part of pretalx itself, making it possible as a plugin is a good starting point.

I also once tried it to implement OIDC login as a plugin but failed due to missing interfaces /signals to "enhance" pretalx accordingly. We started to maintain an internal pretalx fork supporting SSO via OIDC, but I would really like to delete it again :D

rixx:
Sorry for leaving you without a response for so long – I started the process of moving to a new place at the other end of the country pretty much the *day* you opened this PR, and am only now back at regular working life. Just really unfortunate timing.

src/pretalx/common/templates/common/auth.html
```diff
@@ -39,6 +40,9 @@ <h4 class="text-center">{% translate "I already have an account" %}</h4>
                 {{ phrases.base.password_reset_heading }}
             </a>
         {% endif %}
+        {% with get_success_url|default:'' as success_url %}
```
rixx, line 43: Niggle: I’d prefer to pass `success_url` and/or `next_url` as attributes/variables here, instead of calling a method from the template. If you don’t want to handle this, I can fix it after merging the PR.

src/pretalx/person/signals.py
```diff
@@ -8,3 +8,13 @@ def register_speaker_csv_exporter(sender, **kwargs):
...
+
+
+deactivate_user = Signal()
+"""
+This signal is sent out when a user is deactivated (i.e. deleted on the
+frontend).
+
+You will get the user as a keyword argument ``user``. Receivers are expected to
+delete any personal information they might have stored about this user.
+"""
```
rixx, line 20: I'd prefer to call this signal `delete_user`, as the deactivation process is not recoverable, and should be treated like a full deletion by all other parties (we only keep the user object around to show a "deleted user" placeholder in logs, proposals etc).

There is a second method, `User.shred()`, which deletes a user account entirely, which is only possible when they don’t have any proposals or teams. Probably best to call this signal in that method too, as from an auth plugin’s POV, deactivation and actual deletion should be treated identically.

rixx (commented):
Two comments (one niggle you can ignore, one note on user deletion).

I have to admit that it’s been a bit since I’ve given pluggable auth serious thought – just because I haven’t had the time. This approach looks good and has the advantage of simplicity / ease of maintenance on the pretalx side, so I don’t see any problem with it, and am generally happy to merge it.

Before I do that, though, I have some questions (no blockers – I promise we’ll move this forward before the next pretalx release one way or another, and I hope to have a new release before/by the end of February):

Is there anything pretalx can/could/should do to make additional auth backends easier to implement and more uniform to develop? I looked at the approach that [pretix](https://pretix.readthedocs.io/en/latest/development/api/auth.html) takes for pluggable auth, and it looks more rounded / more generous to the plugin developer, but at a glance, I’m not sure if that’s something we could add later – looks like we would screw existing auth plugins like yours by changing the process down the line, which I’d really like to avoid.

adamskrz:
> Sorry for leaving you without a response for so long

No worries, hope your move went well. To be honest, I almost forgot about this PR since getting buried with assignments back at university after Christmas... time to refresh my brain on this a bit.

> Is there anything pretalx can/could/should do to make additional auth backends easier to implement and more uniform to develop?

I actually looked at the way pretix does it too when trying to work out how additional backends should be added in the settings file. It's interesting in how the three types of auth system have predefined flows with some parts already added for the developer, but I'm not sure how much work that actually saves rather than creating more in having to work to pretix's specifications - though it certainly makes it more unified for users. It also appears to be very restrictive about certain things, like only allowing users to have a single auth backend. Either way, I don't think adding more specific functions down the line would screw over plugins like mine unless pretalx was to replace django base auth in the way pretix does, which I don't think would be entirely necessary.

The main thing that would be useful is a simpler way to integrate into pretix's UI - much of this would probably be helpful for non-auth plugins as well. For example, in developing my plugin I think the part I got to is adding a way for users to manage auth connections, which is where the `profile_bottom_html` signal is used. This page is quite confusing to add items to as it uses a completely different set of CSS classes in the two locations it exists (orga vs submission pages) despite having almost the same content.

In terms of auth-specific things, the main thing I can think of is controlling or disabling other auth methods in CfP (at least). Currently, with these signals I can add social auth providers below the login/register form, but I'd rather users don't see the option to use a username/password (or are at least drawn away from it). I suppose this could be done with JS, but that isn't ideal.

Apart from that, the main thing I can think is how auth providers could be handled on a per-event vs instance basis. Currently, I've only allowed instance-wide settings, mostly due to this being the easiest to implement with python-social-auth (as well as avoiding having to decide what shows up on the orga login), but as I've mentioned in the plugin readme this has potential issues where custom domains are used, as some providers (e.g. Microsoft) only allow the registered domains for an OAuth app to authenticate. I tried to get my head around the pretalx custom domain system, but can't work out if it would be feasible to redirect to the main domain for auth redirects then back to the custom one - without further thought on this the entire system could easily cause errors on public instances like pretalx.com. 

I hope that makes some level of sense, and apologies that it's taken me a bit to respond!

src/pretalx/common/templates/common/auth.html
adamskrz: Completely agree with you here, I also think this is very messy. I did have a go at trying to get this to use attributes instead but there were too any layers and forms to follow! More than happy for you to fix it after.

src/pretalx/person/signals.py
adamskrz: The reason I put this here is I was trying to account for the situation where a plugin might add info that should stick around with proposals etc. once a user has been deleted - though on second thoughts I am struggling to come up with a situation where this would be needed but not just linked to a proposal directly. 
I guess if a usecase came up for that, a new `shred_user` signal could be added?

rixx: I think we’d want to call the signal from both `shred_user` and `delete_user` – plugins should handle both more or less the same, though we can add a kwarg like `db_delete=False` so that the plugins know if they can rely on the user object sticking around.

rixx:
> I tried to get my head around the pretalx custom domain system, but can't work out if it would be feasible to redirect to the main domain for auth redirects then back to the custom one - without further thought on this the entire system could easily cause errors on public instances like pretalx.com.

We should be able to do that – pretalx knows its primary domain, and in the middleware, we are already exempting domains from the redirect ([here](https://github.com/pretalx/pretalx/blob/73538e7b7dd715362d22bd8acd330ddd25612822/src/pretalx/common/middleware/domains.py#L23)), and we could do something similar for auth – however, we should decide where those URLs should go and document it, so that plugins know what to put in their own docs / what to choose as URLs …

rixx (commented):
Alright, this is starting to look pretty ready for merging (minus the part I'll fix after the merge).

I think my main open questions are: Do we want to add the domain handling discussed? And: I think it would be nice to have a minimal example of an auth plugin in our docs, like the [Writing an Exporter Plugin](https://docs.pretalx.org/developer/plugins/exporter/) page. Would you be up for that, having just written a plugin like that?

adamskrz (commented):
I've had a look at the domain handling stuff, and I'm not really sure how it could be done. As the session info (from what I understand) is stored per-domain, redirecting to the primary domain to handle auth wouldn't then be able to authenticate the custom domain session. 

I also can't think of an elegant solution for checking if a redirect is allowed from the primary domain or not, especially if providers were to be loaded from the database at runtime (as `django-allauth` allows, for reference) - perhaps this should be left up to the individual plugins to implement if needed...

I've had an attempt at writing some documentation, but haven't had time to set up RST on my machine so I'm not sure if it all works! Is it the sort of thing you were hoping for? Let me know if any of it needs more detail.

rixx (approved):
@adamskrz sorry, I was kinda drowining in the big API rewrite. Let’s just merge this for now – I don’t think I’ll be quite confident enough to install and configure it for a huge platform like pretalx.com, but the best way to find out and fix the rough edges is to have the whole thing in use by people.

Thank you again for the contribution, particularly the addition of the developer docs!

rixx:
Updated docs and changed the `next_url` handling in c1c5a0. The template now passes `success_url` as `next_url`, and any login view should include a `success_url` in the template context.

rixx:
@adamskrz Now that this PR has been merged and released – is there anything needed from upstream to make SSO logins possible, from your point of view? While I don’t think offering SSO logins on pretalx.com will happen any time soon, I’d be very happy to have a documented route for self-hosted instances to point to.

OctopusET:
This fork also has SSO support https://github.com/fossasia/eventyay-talk I hope it helps

PR #1936, opened 2025-01-10
Fix line break after SVG icon

badbadc0ffee:
This PR closes/references issue https://github.com/F30/samaware/issues/5. It does so by wrapping navigation icons in the appropriate `<i class="fa">` tag; the img tag itself already hat the corresponding `class="fa-img"` attribute.

## How has this been tested?
Lacal installation with samaware plugin.


rixx (approved):
Thank you! I don't think anybody is currently using custom icons here other than you, so I'm kinda relieved that it's working at all 😅

Issue #1892, opened 2024-11-18
Allow organisers to mark speakers as 'arrived' on the speakers' details pages


badbadc0ffee:
### Problem and impact

In order to be able to quickly check in speakers when they arrive, this should be possible on the speakers' details pages or even wherever speakers are mentioned, on the speakers tab for a submission for example. The latter would for example help when all speakers for one talk arrive at the same time.

### Proposed solution / feature

Ideally, I'd be able to check in speakers whenever the speaker name appears on my screen.

On the speakers details age, the "Mark speaker as [not] arrived" button could be in to top left, where the Links section is placed on the Submissions details page, next to the "Reset password" and "Send email" buttons.
In the Speakers tab of a submission, the "Mark speaker as [not] arrived" button could be placed nearby the "Send email", "Edit" and "Remove" buttons.

rixx:
Sounds good! Notably, these buttons either should not exist or be very unobtrusive, greyed-out and disabled before the event starts (up to a day before, or so).

F30:
Afaict, this was fixed through #1942 (with a few follow-ups such as 3e7a583 and #2096).

PR #1942, opened 2025-01-20
Organisers can now mark speakers as 'arrived' in more places.

badbadc0ffee:
This PR closes/references issue #1892. It does so by adding the `Mark speaker as arrived`-button on the speakers detail page and on the speakers tab of the submissions pages of accepted submissions.

The button is only shown if the `orga.mark_speakers_arrived` permission is present. Thus it is in particular not shown when the event is not yet running.
No button is shown when the speaker was already marked as attending. To undo the flag, the user list must be used (which is unchanged). This is to avoid cluttering the UI wit additional buttons when they are typically not needed.

The `SpeakerToggleArrived`-view now redirects to the referring page if the referer is present in the HTTP request or to the speaker details page if the referer is not know.

## Checklist

- [X] My change is listed in the ``doc/changelog.rst``.

### DIFF
--- a/src/pretalx/orga/templates/orga/speaker/form.html
+++ b/src/pretalx/orga/templates/orga/speaker/form.html
@@ -29,6 +29,7 @@
 
 {% block content %}
     {% has_perm "orga.change_speaker" request.user form.instance as can_edit_speaker %}
+    {% has_perm "orga.mark_speakers_arrived" request.user request.event as can_mark_speaker %}
     {% has_perm "orga.send_mails" request.user request.event as can_send_mails %}
     <h2 class="d-flex justify-content-between align-items-baseline">
         {% include "orga/includes/user_name.html" with user=form.instance.user lightbox=True %}
@@ -42,6 +43,11 @@
             {{ proposal_title }})
         </span>
         <div class="ml-auto mb-1">
+            {% if can_mark_speaker and accepted_submissions.count and not form.instance.has_arrived %}
+                <a class="btn btn-info" href="{{ form.instance.orga_urls.toggle_arrived }}">
+                    {% translate "Mark speaker as arrived" %}
+                </a>
+            {% endif %}
             <a href="{{ form.instance.orga_urls.password_reset }}" class="btn btn-info flip">{{ phrases.base.password_reset_heading }}</a>
             {% if can_send_mails %}
                 <a class="btn btn-outline-info ml-2" href="{{ request.event.orga_urls.compose_mails_sessions }}?speakers={{ form.instance.user.code }}">
--- a/src/pretalx/orga/templates/orga/speaker/form.html
+++ b/src/pretalx/orga/templates/orga/speaker/form.html
@@ -43,7 +43,7 @@
             {{ proposal_title }})
         </span>
         <div class="ml-auto mb-1">
-            {% if can_mark_speaker and accepted_submissions.count %}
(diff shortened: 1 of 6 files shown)


src/pretalx/orga/templates/orga/speaker/form.html
```diff
@@ -42,6 +43,11 @@ <h2 class="d-flex justify-content-between align-items-baseline">
             {{ proposal_title }})
         </span>
         <div class="ml-auto mb-1">
+            {% if can_mark_speaker and accepted_submissions.count and not form.instance.has_arrived %}
+                <a class="btn btn-info" href="{{ form.instance.orga_urls.toggle_arrived }}">
+                    {% translate "Mark speaker as arrived" %}
```
rixx, line 48: It would probably be good to show the speaker’s `has_arrived` status here if they have arrived already, as otherwise the button *and* information will just … not be there.

This should also be a full toggle button (ie also mark as unarrived), as allowing users to revert an accidental click would be good.

src/pretalx/orga/views/speaker.py
```diff
@@ -265,10 +270,8 @@ def dispatch(self, request, event, code):
             person=self.request.user,
             orga=True,
         )
-        if request.GET.get("from") == "list":
-            return redirect(
-                reverse("orga:speakers.list", kwargs={"event": self.kwargs["event"]})
-            )
+        if referer := request.META.get("HTTP_REFERER"):
+            return redirect(referer)
```
rixx, line 274: I would prefer to allow an explicit set of URLs here rather than redirecting to any header sent. Either the `from=` (with values list/profile/`submission:code`) or the standard `next=` (with the same checks for safe urls as in the login view) would be fine.

rixx (changes requested):

src/pretalx/orga/templates/orga/speaker/list.html
```diff
@@ -56,8 +56,8 @@ <h2>
                             <td class="text-right">
                                 {% if profile.accepted_submission_count %}
                                     {% if can_mark_speaker %}
-                                        <a class="btn btn-sm btn-{{ profile.has_arrived|yesno:"success,info" }}" href="{{ profile.orga_urls.toggle_arrived }}?from=list">
-                                            {% if profile.has_arrived %}
+                                    <a class="btn btn-sm btn-{{ profile.has_arrived|yesno:"success,info" }}" href="{{ profile.orga_urls.toggle_arrived }}?next={{ request.path }}">
```
rixx, line 59:
```suggestion
                                    <a class="btn btn-sm btn-{{ profile.has_arrived|yesno:"success,info" }}" href="{{ profile.orga_urls.toggle_arrived }}?next={{ request.path|urlencode }}">
```

src/pretalx/orga/templates/orga/submission/speakers.html
```diff
@@ -60,6 +61,15 @@ <h5>{% translate "Other proposals by this speaker:" %}</h5>
                 {% endif %}
             </p>
             {% if can_edit_speakers %}
+                {% if can_mark_speaker and submission.state in 'accepted,confirmed' %}
```
rixx, line 64: Would probably make sense to extract this block into a template included in the three templates.

rixx (approved):
Some more comments, mostly for my own reference – I'll be making those small changes after the merge. Thank you for the work!

Issue #1946, opened 2025-01-23
Show message on empty list of access codes


rixx:
When an event has no access codes yet, show an empty table rather than nothing at all.

PR #1977, opened 2025-02-28
Show an empty table if the event does not have access codes

kamo-develop:
This PR closes/references issue #1946.


rixx (commented):
Sorry, I should have been clearer here: I would want there to be a separate message specifying that there are not access codes yet. However, in the meantime, we have reworked and streamlined the list views to use a shared template, so I’m adding the change there.

PR #2005, opened 2025-04-17
fix: move export qrcode image into window

psy:
The qrcode image for exporting the schedule overlaps the window and is only displayed half, as shown in the following screenshot:

image

This PR sets `right: 0px`  on the `absolute` positioned container, moving the qr code into the window. 

## How has this been tested?
I tested only manually in the browser, see screenshot:

image

## Checklist

### DIFF
--- a/src/pretalx/static/common/css/_pretalx.css
+++ b/src/pretalx/static/common/css/_pretalx.css
@@ -7,6 +7,7 @@
   width: 160px;
   height: 160px;
   position: absolute;
+  right: 0px;
   background: white;
   box-shadow: var(--shadow-light);
   svg {


rixx (approved):
thanks!

rixx (commented):
Actually: which browser are you seeing the first screenshot behaviour in? I tried this with dev / pretalx.com, and both Firefox and Chrome showed the QR code positioned like in your second screenshot.

This is because of the block ``.export-qrcode-image { right: 0 }`` that is a few lines below your changes ([here](https://github.com/pretalx/pretalx/blob/c2eb0b618b42d0554ce4b7fa53fc6179e0d6e9c4/src/pretalx/static/common/css/_pretalx.css#L17)), which should handle this already …

psy:
Uhm, maybe the issue is related to the versions of the two pretalx instances i tested (eh, congress). If it's already fixed in a newer version, this PR is obsolete. Sorry for the noise :/

PR #2010, opened 2025-04-22
Allow non-public resources to be used

badbadc0ffee:
This change allows speakers and organizers to mark resources as non-public or private.
Such resources can help reviewers a lot during tzhe review phase but can also be used by a translation crew during an event without making the files and links public in the official schedule.

## How has this been tested?

This has right now only been tested manually.

## Checklist


src/pretalx/cfp/templates/cfp/includes/submission_resources_form.html
```diff
@@ -10,7 +10,7 @@
     <div data-formset-body class="col-md-9">
         {% if can_edit %}
             <div class="alert alert-info flip ml-auto">
-                {% translate "Resources will be publicly visible." %}
+                {% translate "Resources will be publicly visible unless they are markes as non-public." %}
```
rixx, line 13:
```suggestion
                {% translate "Resources will be publicly visible unless they are marked as non-public." %}
```

src/pretalx/submission/models/resource.py
```diff
@@ -32,6 +32,7 @@ class Resource(PretalxModel):
     description = models.CharField(
         null=True, blank=True, max_length=1000, verbose_name=_("Description")
     )
+    private = models.BooleanField(default=False, verbose_name=_("Non-public resource"))
```
rixx, line 35:
```suggestion
    is_private = models.BooleanField(default=False, verbose_name=_("Non-public resource"))
```

I’d prefer the `is_` naming convention for boolean fields (like with `Question.is_public` and `Event.is_public`). I know we don’t do this everywhere, but I feel like it’s less confusing during development.

rixx (commented):
Thanks! I’m all for this feature, so just some niggles:

- I wonder if we should switch the default/wording to be in line with other models: `is_public` (with `True` being the default). This would also make the wording on the form field easier: "Show publicly" is pretty clear, and still carries the implication that otherwise, organisers have access (imo)
- Could you please remove the changes to the translations? Those will lead to merge conflicts as translations change a lot and are annoying to deal with – I’ll regenerate the translations after the merge.
- Likewise, could you remove the migration? There have been other migrations in the meantime (because I was slow to review this, mea culpa) which would conflict; I’ll generate the new migration after the merge.

dest81:
Any progress on this?

badbadc0ffee:
* [X]  Make sure private resources are not included in the API for public users
_resources appear not to be available via the API at all_

 * [ ]  Make sure the schedule export only includes public resources
 * [ ]  On the submission review page, indicate if a review is public or private
 _the review page currently only shows public resources_

 * [ ]  Include private resources in the submission forms

badbadc0ffee:
@dest81, maybe you can contrubute some tests and / or documentation?
The functionality appears to be fine from my point of view.

dest81:
> @dest81, maybe you can contribute some tests and / or documentation? The functionality appears to be fine from my point of view.

Sure, I can work on tests, but will have a time only in 2 weeks

badbadc0ffee:
@dest81: "Any progress on this?" 😉

src/pretalx/cfp/templates/cfp/includes/submission_resources_form.html
```diff
@@ -10,7 +10,7 @@
     <div data-formset-body class="col-md-9">
         {% if can_edit %}
             <div class="alert alert-info flip ml-auto">
-                {% translate "Resources will be publicly visible." %}
+                {% translate "Resources will be publicly visible unless they are marked as non-public." %}
```
rixx, line 13:
```suggestion
                {% translate "Resources will be publicly visible unless they are marked as non-public." %}
                {% translate "Resources will be publicly visible unless marked as non-public." %}
```

src/pretalx/orga/templates/orga/submission/review.html
```diff
@@ -124,6 +124,20 @@
 
                                 {% include "agenda/includes/submission_resource.html" %}
 
+                                {% if not forloop.last %}<br>{% endif %}
+                            {% endfor %}
+                        </div>
+                    </div>
+                {% endif %}
+                {% if submission.private_resources %}
```
rixx, line 132: Do we want the non-public resources listed separately or in the main list and just marked with the lock symbol?

src/pretalx/submission/models/submission.py
```diff
@@ -981,14 +981,33 @@ def score_categories(self):
...
+                models.Q(  # either the resource exists
+                    ~models.Q(resource="")
+                    & models.Q(resource__isnull=False)
+                    & ~models.Q(resource="None")
+                )
+                | models.Q(  # or the link exists
+                    models.Q(link__isnull=False) & ~models.Q(link="")
+                )
+            )
+            & models.Q(is_public=True)
```
rixx, line 994: I think I'd prefer this as a plain `, is_public=True`, and I feel like we should rename `active_resources` to `public_resources` here – or possibly keep `active_resources` as the overarching property, and then make `active_resources` and `private_resources` simple filters on `active_resources`, with the upside of less code duplication aswell.

rixx (changes requested):
At a minimum before merging this, I'd like to see some tests showing that non-public resources are not included in the public API and in the public session page.

Issue #1976, opened 2025-02-27
Don't show default duration for submission categories in CfP


adamskrz:
### Problem and impact

For our event, we are very flexible on durations for submissions, with categories mostly breaking down into sub 15-minute performances and longer, self contained ones. 
In the category dropdown in CfP, the default duration is shown next to the category name in brackets, which is confusing for people submitting as it suggests there isn't flexibility on length.

### Proposed solution / feature

The easiest fix is probably to not render the duration in SubmissionType.__str__() if cfp_require_duration is True, or it could be a configurable setting?

rixx:
I was going to say that I don’t want additional settings, and to use tracks if you’re set on having session types without a duration, but your idea of checking if the duration field is mandatory sounds really cool!

The only reason I didn’t just implemented it instead of typing this out is that this means 1-2 extra db lookups in the string method here, which can get rough on list pages if we don’t find all the pages where we need to prefetch `event` and `event__cfp`. But I'd be happy to merge a PR doing this!

Pradhvan:
I can take this up 👋🏾

Pradhvan:
> Go for it, [@Pradhvan](https://github.com/Pradhvan) !

Thanks rixx, got a bit occupied earlier. Taking a look at it now.

rixx:
@Pradhvan Please find my review on your PR.

Pradhvan:
> @Pradhvan Please find my review on your PR.

Sure, let me check those out.

PR #2036, opened 2025-05-21
Add option to hide duration in submissiontype

Pradhvan:
- Adds a required_duration setting to the CfP duration field
- Modifies the SubmissionType.__str__ method to check this setting
- Adds a UI toggle in the CfP settings form
- Updates templates to display the new setting

This PR closes/references issue #1976. It does so by modifying the SubmissionType.__str__ method so that the user can have a toggle to show or not show duration in the submission type. 

image

## Checklist

### DIFF
--- a/src/pretalx/orga/forms/cfp.py
+++ b/src/pretalx/orga/forms/cfp.py
@@ -12,22 +12,12 @@ from i18nfield.strings import LazyI18nString
 from pretalx.common.forms.fields import ColorField
 from pretalx.common.forms.mixins import I18nHelpText, JsonSubfieldMixin, ReadOnlyFlag
 from pretalx.common.forms.renderers import InlineFormRenderer
-from pretalx.common.forms.widgets import (
-    EnhancedSelect,
-    EnhancedSelectMultiple,
-    HtmlDateInput,
-    HtmlDateTimeInput,
-    TextInputWithAddon,
-)
+from pretalx.common.forms.widgets import (EnhancedSelect, EnhancedSelectMultiple,
+                                          HtmlDateInput, HtmlDateTimeInput,
+                                          TextInputWithAddon)
 from pretalx.common.text.phrases import phrases
-from pretalx.submission.models import (
-    AnswerOption,
-    Question,
-    QuestionVariant,
-    SubmissionType,
-    SubmitterAccessCode,
-    Track,
-)
+from pretalx.submission.models import (AnswerOption, Question, QuestionVariant,
+                                       SubmissionType, SubmitterAccessCode, Track)
 from pretalx.submission.models.cfp import CfP, default_fields
 from pretalx.submission.models.question import QuestionRequired
 
@@ -58,6 +48,13 @@ class CfPSettingsForm(
             "Allow submitters to share a secret link to their proposal with others."
         ),
         required=False,
+    )
+    show_duration_in_submission_type = forms.BooleanField(
+        label=_("Show duration in submission type names"),
+        help_text=_(
+            "If enabled, submission types will display their duration (e.g., 'Talk (30 minutes)'). If disabled, only the name will be shown (e.g., 'Talk')."
+        ),
+        required=True,
     )
 
     def __init__(self, *args, obj, **kwargs):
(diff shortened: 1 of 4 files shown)


src/pretalx/orga/forms/cfp.py
```diff
@@ -12,22 +12,12 @@
...
 from pretalx.common.forms.mixins import I18nHelpText, JsonSubfieldMixin, ReadOnlyFlag
 from pretalx.common.forms.renderers import InlineFormRenderer
-from pretalx.common.forms.widgets import (
-    EnhancedSelect,
-    EnhancedSelectMultiple,
-    HtmlDateInput,
-    HtmlDateTimeInput,
-    TextInputWithAddon,
-)
+from pretalx.common.forms.widgets import (EnhancedSelect, EnhancedSelectMultiple,
```
rixx, line 15: Please don't commit changes that have nothing to do with the Pull Request, as it only leads to merge conflicts / needs to be reverted

src/pretalx/orga/forms/cfp.py
```diff
@@ -59,6 +49,13 @@ class CfPSettingsForm(
         ),
         required=False,
     )
+    show_duration_in_submission_type = forms.BooleanField(
```
rixx, line 52: I think the discussion in #1976 involved **not** adding a separate setting, right? That was the reason why the feature request was accepted, after all. As far as I can tell, you don’t use this field for determining if the duration should be shown, so I’m confused as to why you added it. Could you please explain your reasoning?

src/pretalx/orga/forms/cfp.py
```diff
@@ -140,6 +145,7 @@ class Meta:
             "submission_public_review": "feature_flags",
             "present_multiple_times": "feature_flags",
             "mail_on_new_submission": "mail_settings",
+            "show_duration_in_submission_type": "feature_flags",
```
rixx, line 148: Why did you add this flag here? You added it on the model, not as a feature flag, so why configure it as part of the feature flags JSON field?

src/pretalx/orga/forms/cfp.py
```diff
@@ -130,6 +130,11 @@ def save(self, *args, **kwargs):
             self.instance.cfp.fields[key]["max_length"] = self.cleaned_data.get(
                 f"cfp_{key}_max_length"
             )
+        if "duration" not in self.instance.cfp.fields:
+            self.instance.cfp.fields["duration"] = default_fields()["duration"]
```
rixx, line 134: Could you please explain why you added this code?

src/pretalx/submission/models/cfp.py
```diff
@@ -48,7 +48,10 @@ def default_fields():
         "do_not_record": {"visibility": "optional"},
         "image": {"visibility": "optional"},
         "track": {"visibility": "do_not_ask"},
-        "duration": {"visibility": "do_not_ask"},
+        "duration": {
+            "visibility": "do_not_ask",
+            "required_duration": True,
```
rixx, line 53: This would simultaneously make the duration field **not** visible and **required**, which I don't think is sensible? Depending on what you meant to do, you should change the `visibility` to `optional` or `required` – but, crucially, I don't understand why you are changing the default visibility of the duration field at all. Could you please explain your reasoning behind this change?

src/pretalx/submission/models/type.py
```diff
@@ -75,6 +75,13 @@ def __str__(self) -> str:
         """Used in choice drop downs."""
         if not self.default_duration:
             return str(self.name)
+
+        if hasattr(self, "event") and hasattr(self.event, "cfp"):
+            if not self.event.cfp.fields.get("duration", {}).get(
```
rixx, line 80: Isn’t this logic the wrong way around? Like this, you are only returning the name if the duration field is **not** required, which does not correspond with the original requirements, if I'm not mistaken.

rixx (changes requested):
I have a bunch of questions regarding this PR – you can find them below. Did you test this change (manually or with automated tests)?

rixx:
Implemented this issue in 4f0e795, for reference.

PR #2050, opened 2025-05-29
Correct false warning pretalx.I003 even with redis properly configured

davidmgvaz:
?: (pretalx.I003) You have no Redis server configured, which is strongly recommended in production.
	HINT: https://docs.pretalx.org/administrator/configure/#the-redis-section

This warning is triggered even with Redis properly configured because, along the way, when memcache was removed, REAL_CACHE_USED was not correctly updated.

It does not solve a problem in itself, but it corrects a false warning.


rixx (commented):
Thank you for reporting the problem! I chose a different fix, removing the `REAL_CACHE_USED` variable entirely.

PR #2072, opened 2025-06-24
Add Korean language (ko) support

OctopusET:
This PR adds Korean language support.

## How has this been tested?

image

I recently updated the translation on Weblate. I think the translations should be merged first.

You might noticed that there was 3 bulk update. I reverted first two bulk upload, and I only added new translations when I add the last one, so original translations are preserved now.

Most of translations are machine translated. If you don't like it, I will revert it again. And our team are going to review all translation soon (maybe in this weekend).

I used patch as a reference for this PR https://github.com/pretalx/pretalx/commit/4a62dbc05e2bfa5ea450b71427944acf15da5ed6

Since it doesn't use ko-KR or ko-KP, so I didn't add the test like Japanese one.
I don't think our North Korean comrades will be mad because I used 'ko' code. It's commonly used for most of projects.

## Checklist


rixx (commented):
I believe Korean is at 99%, as per [translate.pretalx.com](https://translate.pretalx.com/projects/pretalx/pretalx/#translations) – however, you’re right that we should add it to the language list. However, there is a bit more to it (at least adding calendar locales), so I’ll do this myself.

PR #2094, opened 2025-08-07
 Allow plugins to inject additional Forms in orga area

badbadc0ffee:
This is a stripped down proposal for https://github.com/pretalx/pretalx/pull/2017.

This PR only includes the wechanism for form extensions.

The `form_signal` has ben defined analoguously to `html_signal`. It can be used to place additional form elements into existing templates. In order to add Forms the `form_signal` triggers sending another signal to plugins which can in tur retunr a django `Form` or a list of forms.

`pretalx.orga.signals.extra_form` is a generic signal that covers a huge number of existing forms - basically all that are based on the `basic_form.html` template and the `OrgaCRUDView`. In addition to the generic signal - where the plugin has to figure out in which context the additional form is needed - signals `speaker_form`, `submission_form`, `review_form` and `mail_form` have been defined to extend the corresponding forms.

It has currently only been tested manually (but rather extensively) using modified versions of the plugins pretalx-signal-demo, pretalx-rt, and pretalx-zammad which have been developed against the previous proposals. The necessary modifications in these plugins are light weight.

As far as I'm aware of, the previous approaches have not yet been adopted by anyone else. I do like this version more than the previous.

### DIFF
--- a/src/pretalx/common/templatetags/form_signal.py
+++ b/src/pretalx/common/templatetags/form_signal.py
@@ -1,0 +1,28 @@
+from django import template
+from django.utils.module_loading import import_string
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def form_signal(context, signal_name: str, **kwargs):
+    """
+    Usage:
+        {% form_signal "path.to.signal" argument="value" ...  as extra_forms %}
+        {% for form in extra_forms %}
+            {{ form }}
+        {% endfor %}
+    """
+    signal = import_string(signal_name)
+    request = kwargs.pop("request", context.get("request"))
+    sender = kwargs.pop("sender", request.event)
+    forms = []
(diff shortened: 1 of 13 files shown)


badbadc0ffee:
I have added a change to the `out_form.html` for e-mails:
The read-only version of the form (used for sent mails) shall still use the regular `{{ form }}` to render forms provided by a plugin - not render the individual fields on its own. Otherwise, the plugin looses control over the visual represenation.
I have also slightly adapted the rendering of the read only elements of the mail to have a more uniform layout and look of the existing elements and parts added by plugins.

badbadc0ffee:
> [!WARNING]
> Found a problem with `Views` that are not related to an `Event` (i.e. `TeamView`).

The `form_signal` implementation expected to always find an event in the request.
It shoud set the `sender` to `None` for views like the `TeamView`.

PR #2096, opened 2025-08-10
Fix & improve marking speakers as arrived

F30:
This PR can be seen as a follow-up to #1942. It improves various aspects of marking speakers as arrived:

- Fix typo that prevented marking speakers as arrived from their orga detail page
- Restore usage of "mark_speakers_arrived" include in orga speakers list (Had originally been introduced in 3e7a583, apparently got lost in d0617bd1d.)
- Switch marking speakers as arrived to POST request (It alters data, so it shouldn't use GET.)

## How has this been tested?

Ran and updated tests, checked the templates manually.

## Checklist

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,6 +10,7 @@ Chris Wolf
 David Ellis
 Dominik Helle
 Emmanouil Kampitakis
+Felix Dreissig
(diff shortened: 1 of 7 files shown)


rixx (approved):
Thank you!

PR #2097, opened 2025-08-10
Correctly store freshly created draft proposals

badbadc0ffee:
When a user created a draft proposal via the CfP flow, it was initially saved in the database as `submitted` and then immediately updated to the `draft` state.

Code listening to `django.db.models.signals` would have received two `pre_save` and two `post_save` signals — first with the incorrect `submitted` state, then with the correct `draft` state.

Since draft proposals should not be visible (even to organizers), saving them in an incorrect state should be avoided.

## How was this found and tested?

This issue was discovered while working on the RT plugin, which had begun creating tickets even for draft proposals.  
I tested the fix manually and kept the change minimal. I considered changing the default from `submitted` to `draft`, but that would have required a larger changeset to ensure all other submissions were still created as `submitted` — which remains the preferred default in most cases.

## Checklist

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -3,6 +3,7 @@
 Release Notes
 =============
 
+- :bug:`cfp` Draft proposals were created and saved as submitted instead of draft, then changed to draft and saved again. This has been fixed to prevent the submitted state from appearing in the database, even briefly.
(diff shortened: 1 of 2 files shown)


rixx (approved):
Thank you!

PR #2099, opened 2025-08-11
Add shortcut URLs

badbadc0ffee:
This PR adds shortcut URLs of the form ~`/goto/<code>`~ `/redirect/<code>`.

The idea came up while writing the RT plugin. RT allows custom fields that can automatically create links — in this case, back to pretalx. Unfortunately, the link pattern cannot vary by queue in RT (we have one queue per event in pretalx), so I could only create working links to one pretalx event at a time.

The proposed shortcut URLs work independently of the event. Codes are unique across the entire pretalx installation, not just within a single event, so the corresponding event for a code can be determined easily.

The implementation first tries to find a matching submission, then a matching person.  
If the request is made in a browser by an authenticated pretalx user with organizer permissions, the link redirects to the organizer area. If the user is anonymous or lacks organizer permissions, it redirects to the public page.

## How has this been tested?
Only manual tests have been made. Very similar code (without support for speaker codes) has been used during the last congress wir the RT plugin.

## Checklist

- [x] I have updated the documentation (minor fix for changed permissions).

### DIFF
--- a/src/pretalx/common/views/shortlink.py
+++ b/src/pretalx/common/views/shortlink.py
@@ -1,0 +1,31 @@
+from contextlib import suppress
+
+from django.http import Http404, HttpResponseRedirect
+from django_scopes import scopes_disabled
+
+from pretalx.person.models.user import User
+from pretalx.submission.models.submission import Submission
+
+
+@scopes_disabled()
+def shortlink_view(request, code, *args, **kwargs):
+    with suppress(Submission.DoesNotExist):
+        submission = Submission.objects.select_related("event").get(code=code)
+        if request.user.has_perm("submission.orga_list_submission", submission):
+            return HttpResponseRedirect(submission.orga_urls.base)
+        if request.user.has_perm("submission.view_public_submission", submission):
+            return HttpResponseRedirect(submission.urls.public)
+    with suppress(User.DoesNotExist):
+        user = User.objects.get(code=code)
+        if request.user.has_perm("person.administrator_user", user):
+            return HttpResponseRedirect(user.orga_urls.admin)
+        profiles = user.profiles.select_related("event").order_by("-created")
+        for profile in profiles:
+            if request.user.has_perm("person.orga_list_speakerprofile", profile):
+                return HttpResponseRedirect(profile.orga_urls.base)
+        if profile := profiles.first():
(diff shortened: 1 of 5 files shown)


src/pretalx/urls.py
```diff
@@ -25,6 +25,7 @@
     path("orga/", include("pretalx.orga.urls", namespace="orga")),
     path("api/", include("pretalx.api.urls", namespace="api")),
     path("redirect/", redirect_view, name="redirect"),
+    path("goto/<code>", shortlink_view, name="shortlink"),
```
rixx, line 28: I wonder if we should move this endpoint to `/redirect/goto/` or to `/api/goto/` instead – any top-level URL path restricts the choice of event shortcodes (and has the risk of an existing event on some instance conflicting with the chosen URL).

rixx (commented):
Good idea – added some minor notes and remarks, but happy to merge once those are done.

src/pretalx/urls.py
badbadc0ffee: `redirect/<code>` appears to be just fine - that is not in conflict with the `/redirect/?url=...` URLs

src/pretalx/common/views/shortlink.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.http import HttpResponseRedirect
+from django_scopes import scopes_disabled
+
+from pretalx.person.models.user import User
+from pretalx.submission.models.submission import Submission
+
+
+@scopes_disabled()
+def shortlink_view(request, code, *args, **kwargs):
```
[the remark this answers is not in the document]
badbadc0ffee: The 404 for an event that is not publicly visible was actually intentional: you'd e.g. also get a 404 if you tried to open a draft through it's code in the orga area.

badbadc0ffee:
Now I got you.
https://cfp.cccv.de/redirect/PP8N37 redirects to https://cfp.cccv.de/38c3-community-stages/talk/PP8N37/ which shows a 404 but reveals that PP8N37 was submitted to 38c3-community-stages (as a test in that case) but is not public (because it wasn't accepted).
This should indeed be avoided!

rixx:
Merged manually because GitHub was being annoying.

PR #2136, opened 2025-09-26
Order teams by team name instead of ID.

badbadc0ffee:
This is only a cosmetic change and minor enough that I only tested it manually.

I’d prefer sorting Teams by name (while keeping Teams with permissions for all Events at the top) instead of by ID.
This way, users can control the order of their Teams by choosing appropriate names, without the overhead of using an OrderedModel — which I find unnecessary.


rixx (approved):
thank you!

Issue #2147, opened 2025-10-06
Submission form: no validation when room or start date is not set


dest81:
### Problem and impact

When I set only the room (or only the start date) for a confirmed submission in the orga UI and save, form silently ignores changes. Looking [into code](https://github.com/pretalx/pretalx/blob/main/src/pretalx/orga/forms/submission.py#L166) I can see that slot can be saved only when all 3 fields are set.

### Expected behaviour

I would expect a validation message saying that slot related fields are missing. 

### Steps to reproduce

1.
2.
3.
4.

PR #2151, opened 2025-10-08
add validation for room and start date and time

badbadc0ffee:
This PR closes issue #2147. It does so by adding validation functions for room and start data.
> [!NOTE]
> When both room and start date are cleared, _all_ slots will be deleted.

## How has this been tested?
I have only manually tested that validation errors appear when only the room or start field is set, and that slots can be set or cleared when consistent data is entered, or when both fields are empty.

## Checklist

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -3,6 +3,7 @@
 Release Notes
 =============
 
+- :bug:`orga:submission,2147` The room, start and end times of submissions can now be edited.
 - :feature:`dev` Plugins can now inject additional form elements in the organisers area with the ``form_signal`` similar to ``html_signal``.
 - :feature:`cfp` Uploaded resources can now be marked as private (available only to organisers).
 - :feature:`cfp` Organisers can now define an optional opening date before which no submissions can be made.


src/pretalx/orga/forms/submission.py
```diff
@@ -150,6 +150,22 @@ def clean(self):
             )
         return data
 
+    def clean_room(self):
```
rixx, line 153: Please move these checks to the `clean()` method, as the `clean_*` methods are not guaranteed to have access to all data yet. Validation that impacts multiple fields is [best done in `clean()`](https://docs.djangoproject.com/en/dev/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other).

src/pretalx/orga/forms/submission.py
```diff
@@ -181,6 +197,17 @@ def save(self, *args, **kwargs):
             task_update_unreleased_schedule_changes.apply_async(
                 kwargs={"event": self.event.slug}
             )
+        if (
+            instance.state in SubmissionStates.accepted_states
+            and not self.cleaned_data.get("room")
+            and not self.cleaned_data.get("start")
```
rixx, line 203: As the form only passes validation if both `room` and `start` are present (or neither), checking for one here should be enough.

src/pretalx/orga/forms/submission.py
```diff
@@ -181,6 +197,17 @@ def save(self, *args, **kwargs):
             task_update_unreleased_schedule_changes.apply_async(
                 kwargs={"event": self.event.slug}
             )
+        if (
+            instance.state in SubmissionStates.accepted_states
```
rixx, line 201: We should drop this check and go through with the deletion in any case. It's a bit wasteful, admittedly, but this way the logic here won't grow out of sync with the central logic regarding which proposals should have slots. Case in point, a while ago we changed the logic so that submissions that are *pending accepted* also have slots. Those proposals would not be covered by the check above.

rixx (changes requested):
Good catch, thanks! Some minor comments, looking good otherwise.

Issue #2152, opened 2025-10-09
Make created API tokens easier to recognize


badbadc0ffee:
### Problem and impact

When I created the first API key, I didn't realize that the code I needed was already shown in the success message — and that I wouldn't see it ever again.

Just recently, someone else had the exact same experience — I am probably not the only one who didn't actually read the text in the success box.

I'd propose changing the color and icon of that message box to make it more likely that users spot the generated code on their first try.

### Proposed solution / feature

I propose to use extra tags for the message and a definition for a different icon in the message box.

### Context or examples

<img width="1877" height="244" alt="Image" src="https://github.com/user-attachments/assets/4e03fc11-b921-4860-abe6-9a495f730c0c" />

PR #2153, opened 2025-10-09
Show API key message in different color

badbadc0ffee:
This PR is a proposal for #2152.

The success message takes the color of an info message to separate it from typical success messages, which some (including myself) get so used to that they ignore them. To make it even more obvious that this is not a usual 'all fine, go ahead' message, it also shows a different icon.

## Checklist

- [X] My change is listed in the ``doc/changelog.rst``.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -6,7 +6,15 @@
 Release Notes
 =============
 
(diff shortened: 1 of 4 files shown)


rixx (approved):
Thank you!

Issue #2155, opened 2025-10-10
forced pagination limit in api v1 breaks automaticity


erdgeist:
### Problem and impact

When using the api REST endpoint to retrieve all submissions or speakers, being forced to get multiple fragments of the list in multiple requests yields unstable and resulting dupes but also missing entries.

This behaviour comes with an arbitrary limit of 50 entries per page with the stated goal of avoiding performance issues.

### Expected behaviour

The API endpoints should not enforce arbitrary limits. If that is unacceptable for all pretalx instances, requiring a global instance setting combined with a ?hurt_me_plenty=true parameter is fine.

### Steps to reproduce

1.
2.
3.
4.

### Version

API v1

PR #2161, opened 2025-10-14
Also allow changes to the page size limitation for paginated lists.

badbadc0ffee:
This PR adds to issue #2155.

I'd like to propose to also allow overriding the limit for paginated lists.
The performance impact should be in the same ballpark as for the API pagination - I am unsure if a separate setting is needed or the same limit should simply be appied to API and lists.
I didn't know in which section of the config the option would fit best...

## How has this been tested?
Manually. With a _huge_ DemoCon...

## Checklist


OctopusET:
This would be useful

rixx (commented):
I thought about this a bit and in the end decided against a separate limit, and instead moved the existing setting from the [api] section to the [site] section, and updated the docs to indicate that it will apply to both the API and the UI. I used your implementation otherwise (co-author credits are in commit metadata)

Issue #2167, opened 2025-10-22
Frontpage header pattern cannot be (re-)set to 'plain'


badbadc0ffee:
### Problem and impact

With the latest `main` version, it is not possible to reset the frontpage header pattern to `plain`.

The error message is
_Select a valid choice. on is not one of the available choices._

PR #2168, opened 2025-10-22
don't html-minimize event settings form

badbadc0ffee:
mimimizing the form strips the value="" part of the radio button for the plain header pattern

This PR closes issue #2167. It does so by disabling the HTML minimizer for the EventDetails view which was broken, because the empty `value=""` part of the radio button for the plain header pattern was inadvertently stripped.

## How has this been tested?
Tested manually in a dev. environment.

## Checklist


rixx (approved):
Thank you!

PR #2171, opened 2025-10-24
User mails can not be expanded in cfp form

badbadc0ffee:
I've just seen that mails cannot be expanded in the user mails view in the CfP because the JS was missing.

## How has this been tested?

almost not at all yet...

## Checklist


rixx (approved):
Thank you!

PR #2195, opened 2025-11-22
Don't show auto-created template copies in template table

badbadc0ffee:
Auto created copies of used custom mail templates have inadvertently been shown in the templates table.


rixx (approved):
Thank you!

PR #2215, opened 2025-12-04
update documentation for API access to exporters

badbadc0ffee:
Minor proposal for the API documentation / example.


rixx (commented):
I'm generally cautious about changing translations, as it will mean that all community translators will have to review the changed string. I'll pull in the addition of the text to the speaker export page and the fixed url on the schedule export page

PR #2218, opened 2025-12-05
Add feedback API

rixx:


src/pretalx/api/serializers/feedback.py
```diff
@@ -0,0 +1,65 @@
...
+        allow_null=True,
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.event:
+            self.fields["submission"].queryset = self.event.talks
+            self.fields["speaker"].queryset = self.event.speakers
+
+    def validate_submission(self, value):
```
badbadc0ffee: maybe also validate the speaker so only speakers of the submission can be used - or is this covered by other mechanisms?

src/pretalx/api/serializers/feedback.py
```diff
@@ -0,0 +1,65 @@
...
+        ]
+
+
+@register_serializer(versions=CURRENT_VERSIONS)
+class FeedbackSerializer(FeedbackWriteSerializer):
+    submission = SlugRelatedField(slug_field="code", read_only=True, source="talk")
+    speaker = SlugRelatedField(slug_field="code", read_only=True)
+
+    class Meta(FeedbackWriteSerializer.Meta):
+        expandable_fields = {
```
badbadc0ffee: maybe limit the expandable fields to avoid leaking information if only feedback should be allowed

badbadc0ffee (approved):

PR #2232, opened 2025-12-13
Add hide until event day option for session resources

OctopusET:
This PR adds a "hide until event day" option for session resources, allowing speakers to upload materials in advance while keeping them hidden from attendees until their session starts.

## How has this been tested?
  - Manually tested creating resources with the hide option enabled/disabled
  - Verified resources are hidden before session start and visible after
  - Added automated tests for the visibility logic
  - Tested API responses respect the visibility settings

## Checklist


src/pretalx/submission/models/submission.py
```diff
@@ -1041,6 +1041,11 @@ def private_resources(self):
     def public_resources(self):
         return self.active_resources.filter(is_public=True)
 
+    @cached_property
+    def available_public_resources(self):
```
rixx: I don't think we want a new property here, and instead use this logic in `public resources`. This makes sure that any plugin relying on `public_resources` will continue to only receive the correctly filtered list.

OctopusET:
I just sent it

src/pretalx/submission/models/resource.py
```diff
@@ -65,3 +73,25 @@ def filename(self):
...
+        earliest_slot = (
+            self.submission.slots.filter(start__isnull=False).order_by("start").first()
+        )
+
+        if not earliest_slot or not earliest_slot.start:
+            # No scheduled slot yet, so resource is not available
+            return False
+
+        # Resource is available if current time is on or after the slot start
+        return now() >= earliest_slot.start
```
rixx: The user is told that the resource will start being visible on the day of the session, but per this code, it only starts being visible at the *time* of the session.

src/pretalx/submission/models/resource.py
```diff
@@ -40,6 +41,13 @@ class Resource(PretalxModel):
     is_public = models.BooleanField(
         default=True, verbose_name=_("Publicly visible resource")
     )
+    hide_until_event_day = models.BooleanField(
```
rixx: I don't think a new field is the right choice here, in terms of data-modeling. What would it mean for a resource to be set to `is_public=False` but `hide_until_event_day=True`? Would the user expect that the resource would be hidden from the organisers until the event day?

rixx (changes requested):

src/tests/submission/test_resource_model.py
```diff
@@ -0,0 +1,76 @@
+# SPDX-FileCopyrightText: 2025-present Sungjoon Moon
```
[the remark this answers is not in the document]
OctopusET: I just followed other files

OctopusET:
Hi, sorry for late reply.

I’ve reconsidered this feature, and I think it might make it harder for attendees to access the materials. Some people might want to read them in advance. Also, if this option is visible, more people will likely enable it. So I feel this feature might be an anti-pattern. What do you think? If you still think it's worth having, I'll go ahead and fix the issues in the current PR.

rixx:
@OctopusET I've been similarly torn about this, and I think you are right: People can always decide to change their slides later on, or only upload after the fact manually. If organisers are concerned that speakers will forget, they can request the slides via a custom field that is private, and turn it public later on (e.g. after vetting it, or the da of the event).

I think it's best not to have this feature in core, even though it's sad that you put your time and thoughts into this, only for the merge to not take place. I really appreciate your thoughtfulness (and the fact that you contributed – taking a decision against a feature and arguing about it is as much a good contribution as getting a feature merged).

PR #2240, opened 2025-12-19
Slightly improve German translation

michikommader:
This PR improves German translation and language style in a few places.

## How has this been tested?

Not tested. I  expect these minor changes in characters/words to blend-in neatly.

## Checklist


rixx (commented):
First off, please use translate.pretalx.com to contribute to translations.

Secondly: You’re changing German imperative verbs to use a form that is no more or less correct than the one currently in use. There is a case to be made that „lade“ is preferable to „lad“ according to some obscure rules, but in cases like „geh“ vs „gehe“, I do not believe that one form is strictly preferable, so I’d prefer to do without this change.
If these changes are important to you, you have the option of using a language provider plugin to override these strings.

PR #2241, opened 2025-12-19
Add editable API tokens

rknall:
## Summary
- Allow editing existing API tokens (name, events, expiration, permissions)
- Token value remains unchanged during edits
- Fix validation display for required enhanced select fields

## Test plan
- [x] Create a new token and verify it works
- [x] Edit an existing token's name, events, and permissions
- [x] Verify token value doesn't change after editing
- [x] Test validation error displays when saving without events
- [x] Verify expired tokens cannot be edited
- [x] Verify other users' tokens cannot be edited
- [x] All existing tests pass

### DIFF
--- a/src/pretalx/common/log_display.py
+++ b/src/pretalx/common/log_display.py
@@ -140,8 +140,10 @@ LOG_NAMES = {
     "pretalx.speaker_information.create": _("A speaker information note was added."),
     "pretalx.speaker_information.update": _("A speaker information note was modified."),
     "pretalx.speaker_information.delete": _("A speaker information note was deleted."),
+    "pretalx.user.token.create": _("An API token was created."),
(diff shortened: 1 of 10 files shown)


rixx:
I'm not convinced that we do want editable API tokens. I'll have to research the security implications first.

rixx:
No – I attended 39C3 as mentioned previously, and at the tail end caught influenza, which I am only just recovered from.

rixx:
After giving this some thought, I decided that only very bounded edits are okay for now: name and expiration cannot be freely edited (though you can still expire tokens with the click of a button), API version is updated via the existing mechanism, and users can now update the set of included events for tokens.

Thank you for discussing this and convincing me that this would be useful!

PR #2246, opened 2026-01-02
fix(cfp wizard): handle ValueError when invalid GET params are passed

dest81:
This fix ensures graceful handling of AttributeError caused by invalid submission_type or track values in GET parameters.
```
Internal Server Error: /democonf/submit/paPgsd/info/
ValueError at /democonf/submit/paPgsd/info/
invalid literal for int() with base 10: 'anystring'

Request Method: GET
Request URL: http://localhost/democonf/submit/paPgsd/info/?submission_type=any_string
```

## How has this been tested?
I have fixed the test that suppose to catch the error. Test was incorrect and  did test empty values.

## Checklist

### DIFF
--- a/src/pretalx/cfp/flow.py
+++ b/src/pretalx/cfp/flow.py
@@ -388,7 +388,7 @@ class InfoStep(DedraftMixin, FormFlowStep):
         for field, model in (("submission_type", SubmissionType), ("track", Track)):
             request_value = self.request.GET.get(field)
             if request_value:
-                with suppress(AttributeError, TypeError):
+                with suppress(AttributeError, ValueError, TypeError):
                     pk = int(request_value.split("-")[0])
                     obj = model.objects.filter(event=self.request.event, pk=pk).first()
                     if obj:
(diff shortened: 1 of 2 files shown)


src/pretalx/cfp/flow.py
```diff
@@ -388,7 +388,7 @@ def get_form_initial(self):
         for field, model in (("submission_type", SubmissionType), ("track", Track)):
             request_value = self.request.GET.get(field)
             if request_value:
-                with suppress(AttributeError, TypeError):
+                with suppress(AttributeError, ValueError, TypeError):
                     pk = int(request_value.split("-")[0])
```
dest81: **comment:** I am not sure what is the purpose of `.split("-")` here

badbadc0ffee: The `split()` is used to support values in the form of `1-talk` and `2-workshop` where the text part is ignored and the integer indicates the PK for the session type.
`any_text` without a leading integer does not make a lot of sense  - and personally I prefer the error 500 in that case over silently ignoring the error.

PR #2254, opened 2026-01-08
Allow filtering the schedule before exporting

johanvdw:
This is a small proposal for adding a filter filed to the schedule exporter.

This allows plugins to reuse much of the logic,  but adding a custom filter, eg only a specific track, day or whatever other field they choose.

## How has this been tested?
I have not yet added a unit test, but I will do so if this seems useful.  Idem for documentation.

We have this change in our branch for FOSDEM (we have a field to hide tracks from the website)

## Checklist

### DIFF
--- a/src/pretalx/schedule/exporters.py
+++ b/src/pretalx/schedule/exporters.py
@@ -19,11 +19,12 @@ from pretalx.schedule.ical import get_slots_ical
 
 
 class ScheduleData(BaseExporter):
-    def __init__(self, event, schedule=None, with_accepted=False, with_breaks=False):
+    def __init__(self, event, schedule=None, with_accepted=False, with_breaks=False, filter=None):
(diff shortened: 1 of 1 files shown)


rixx (changes requested):
`filter` is a built-in function in Python, and I'd prefer if we didn't shadow it (even though it's rarely used). How about `filter_q` or something like this (which would also clarify the type a bit)?

rixx:
I looked at this again, and apart from the filter name, the biggest problem is that the way we are using exporters currently is hacky and bad: we instantiate them without the schedule, then inject it later. We do this to avoid instantiating it twice (once without a schedule for access checks, then with for data access). Terrible pattern. I'll look into untangling this into more normal use, and I'll add a `talks_filter` kwarg defaulting to empty `Q()` while I'm at it.

PR #2277, opened 2026-02-02
Fix the ValueError when sending test email in the admin

harriebird:
The Send test email in the Administration information page causes a ValueError when clicked. This error is caused by expecting a name value coming from the `settings.ADMINS`. The email in the `logging` section inside the configuration file only contains email or comma-separated emails.

A screenshot of the error is shown below:

screenshot

## How has this been tested?
This fix was tested manually and automatically. The value of `settings.ADMINS` inside the `test_test_mail_sends_email` test was changed to a value without the name.

This is the result of the manual test:

Screenshot From 2026-02-02 18-33-09

## Checklist

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -6,6 +6,7 @@
 Release Notes
 =============
 
+- :bug:`orga` Sending a test email in the Administration Information page causes a ValueError.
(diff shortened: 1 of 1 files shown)


rixx (commented):
thank you for reporting this problem! As I wanted to make a related change, I just fixed this myself, but I appreciate the PR all the same.

PR #2349, opened 2026-03-21
feat: add tab as CSV data delimiter option

### LINKED ISSUE #2265: Add "tab" as delimiter in CSV export
### Problem and impact

Related to https://github.com/pretalx/pretalx/issues/2264, I'd like the option to choose "tab" as a delimiter. Both newline and comma come with their own set of problems (as does tab and CSV in general, I know... ;) and tab is slightly less bad (IMO)

### Proposed solution / feature

_No response_

### Context or examples

_No response_

mvanhorn:
## Summary

Adds "Tab" as a third option for the CSV data delimiter, alongside "Newline" and "Comma".

## Why this matters

Per #2265, both newline and comma delimiters have drawbacks when data contains those characters. Tab is a common alternative that avoids issues with commas in text fields and newlines in multi-line content.

## Changes

- `src/pretalx/orga/forms/export.py:32`: Added `("tab", _("Tab"))` to the `data_delimiter` choices
- `src/pretalx/orga/forms/export.py:145`: Added `"tab": "\t"` to the `delimiters` dict in `csv_export()`

A previous attempt (PR #2292) was closed without merge. This PR takes the same minimal approach - two lines changed.

## Testing

The existing delimiter tests in `test_orga_views_speaker.py` and `test_orga_views_schedule.py` cover the delimiter selection flow. Tab follows the same code path as newline and comma.

Fixes #2265

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/pretalx/orga/forms/export.py
+++ b/src/pretalx/orga/forms/export.py
@@ -29,7 +29,7 @@ class ExportForm(forms.Form):
         help_text=_(
             "How do you want to separate data within a single cell (for example, multiple speakers in one session/multiple sessions for one speaker)?"
         ),
-        choices=(("newline", _("Newline")), ("comma", _("Comma"))),
+        choices=(("newline", _("Newline")), ("comma", _("Comma")), ("tab", _("Tab"))),
         widget=forms.RadioSelect,
     )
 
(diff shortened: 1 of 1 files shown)


rixx (commented):
Repeating the question that is still open in the linked issue before we discuss merging this issue: Could you explain under which circumstances you would want this?

My reason for asking this is that the deliminter chosen here is not, after all, the delimiter between the CSV fields, which is always a comma. It is the delimiter between sub-fields inside the data, e.g. if you have multiple speakers in a session line, you would separate the speakers by this character. I feel like using a tab here is both unusual and misleading because users will be even more likely to think that they are picking the main CSV delimiter.

rixx:
We are currently migrating to a new export mechanism and will have to revisit sub-field divisions at that point.

PR #2394, opened 2026-04-15
Do not allow creation of API tokens with no permissions

rixx:


rixx:

PR #2396, opened 2026-04-15
Disable email address updates via the API

rixx:


rixx:

PR #2397, opened 2026-04-15
Use POST for all action endpoints

rixx:


rixx:

PR #2467, opened 2026-05-19
Changed 'str' to 'string' in API schema

JonathanOrd:
Changed the type of multiple properties from 'str' to 'string' in `doc/api/schema.yml`, as they were inconsistently named. This issue was noticed when parsing the schema file.

### DIFF
--- a/doc/api/schema.yml
+++ b/doc/api/schema.yml
@@ -4910,7 +4910,7 @@ components:
         answer:
           type: string
         answer_file:
-          type: str
+          type: string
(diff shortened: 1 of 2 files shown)


rixx (commented):
Thank you for reporting this problem. However, as the schema.yml file is generated from the API code, modifying the file will not actually resolve the problem, because when it is generated the next time, the `type: str` fields will be injected again. I will investigate fixing this on the code side.

rixx:
Fixed in 5e2f480, thank you again for pointing this out!

Issue #1692, opened 2024-02-20
Poster session functionality


dionzand:
### Problem and impact

Currently there is no dedicated functionality/plugin for organizing a poster session. Scheduling a poster session as separate "talks" is not ideal, since poster sessions are in parallel.

### Proposed solution / feature

It would be nice to have dedicated functionality to schedule a poster session, where multiple "talks" can occur simultaneously. Or to have a "room" that is dedicated to a poster session, and you can even maybe create a floorplan.

### Context or examples

Poster sessions are often an important part of (scientific) conferences. Maybe this functionality can even be expanded to "exhibition stands", which are also often part of large scientific conferences.

rixx:
This would definitely be cool to have in pretalx, but would also a fair bit of effort to add. Whoever implements this will have to consider the following:

- How to model this on the session side: Is the "poster session" a functionality of the submission object, or of its session type, or something else? I believe it ought to be the session type, as it's already used to determine session duration, and fits in there.
- How to model this on the schedule / talk slot side. Are the poster session parts regular talk slots (which would be good for a lot of interfaces, like showing a session's location and date, but bad because it breaks expectations e.g. re: uniqueness of talk slot IDs), or is it a new object type like a ChildTalkSlot? Possibly a TalkSlot with type="dependent" could work, and we just never render those in APIs that we would break this way 
- Make sure exports and APIs still work. APIs in particular do not expect nested sessions, so figuring this out in a backwards-compatible way that isn't painful to use going forward will take some doing
- Consider permissions – with how pretalx works, reviewers could have access to only some posters, but not the superordinate session, and we'd have to make sure that's honoured
- UI considerations: how do we show a poster session in the schedule? The session detail page should not be a problem, but the overall schedule will prove challenging
- UI considerations: how do posters get added to and moved between poster sessions? Do we have this functionality in the schedule editor, or just in the individual poster form? Can you then re-schedule a poster? etc etc
- Make sure changes show correctly in the schedule changelog, which means that TalkSlots will have to have some knowledge of their child slots, and can't just directly refer to a session

I'm sure this is not all of it, but that's what came to mind lately. Happy to work on this on a sponsored/commissioned basis, or review a PR, as I don't think I'll find the time to prioritise this any time soon otherwise.

johanvdw:
At FOSDEM we have a call for stands, which has very similar requirements. We have used pretalx for this the first time this year (as a separate conference), and we were happy that we could reuse the same review interface as for talks.

[...]

PR #2673, opened 2026-08-18
Poster sessions

RemDelaporteMathurin:
This PR closes/references issue #1692.

@rixx I used Claude to get this so I'm happy to consider this PR as an idea rather than something to be merged if you don't want AI-assisted PRs.

## The problem

pretalx assumes a room holds one session at a time. Poster sessions work the
other way around: dozens of posters are presented side by side, in one place, at
the same time. The same is true of exhibition stands (#1692).

Today that is not schedulable. Placing 50 posters in one room at one time means
50 overlap warnings before every release, and a public schedule that paints all
50 cards on top of each other in a single grid column.

## The change

Rooms gain a `parallel_sessions` flag, set in the room's settings form. In a
parallel room:

- **Release warnings no longer report room overlaps**, since overlapping is the
  point. Speaker overlaps still warn — one person cannot present at two boards
  at once.
- **The schedule editor accepts overlapping drops**, and collects concurrent
  sessions into a single cell holding a scrollable list of compact cards, rather
  than one card per lane. Room columns keep their uniform width, so the layout
  scales the same at 5 posters and at 50.
- **The public grid** does the same: sessions sharing a time slot become one
  cell listing them all. The list view already handled parallel sessions and is
  unchanged.

Sessions are grouped by *overlap*, not by identical start and end, so a poster
that finishes half an hour early joins the cell it overlaps instead of becoming
a second cell painted over the first. When a cell's sessions do not all share
its span, the ones that differ show their own times; the rest stay silent rather
than repeating the cell heading.

Poster file uploads need no new code: proposals already carry resources.

## Also in here

- **Drop targeting in the editor no longer assumes uniform column widths.**
  `updateHoverSlice` used to derive the room index by dividing the pointer
  offset by the first column's width; it now hit-tests the room columns
  directly and clears the hover when the pointer is outside all of them. No
  behaviour change today, since all columns are again equal width, but the
  assumption was silently load-bearing.
- **An unrelated build fix**: `~` alias resolution in the frontend Vite config
  was broken on Windows (`path.sep` matching against importer paths that always
  arrive POSIX-separated). Happy to split this into its own PR if preferred.

## Testing

`just test-parallel` — 9332 passed, 2 skipped. New coverage:

- `tests/schedule/domain/test_warnings.py` — room overlaps suppressed in
  parallel rooms, speaker overlaps still raised, and the query count held flat
  now that `room` is `select_related` in the overlap scan
- `tests/schedule/interfaces/forms/test_room.py` — the form field
- `tests/schedule/interfaces/test_widget.py` — `parallel` in the widget payload,
  including an n+1 test parametrized over 1 and 3 sessions
- `tests/api/serializers/test_room.py`, `tests/schedule/models/test_room.py`

## Notes

- Printing a parallel cell expands it to its full content rather than clipping
  to the slot height, so a printed schedule lists every poster.
- An earlier revision of this branch also added a `board_number` on scheduled
  sessions, for identifying the poster board within the room. It is deliberately
  not part of this PR — the flag alone makes poster sessions schedulable, and
  board numbering deserves its own design discussion (in particular where
  numbers are assigned, and whether they should be generated).

## Screenshots

Room editor with parallel session checkbox:
room_editor

Backend Schedule View:
poster_room_backend

Frontend Schedule View:
poster_room_frontend


rixx:
In my opinion, AI is at most a tool, but cannot replace knowledge of a code base, experience and judgement. In other words, without guidance and review, an AI-agent driven implementation is of a quality where closing it without looking at it and rewriting it from scratch costs less time than trying to salvage it.

In case that sounds like hyperbole, in the case of the current PR:

- It ignores my mention of a required underlying data model change.
- It implements a version of the feature that we were not looking to build: The original issue contained several options, but further discussion (and my [list of questions](https://github.com/pretalx/pretalx/issues/1692#issuecomment-2426388892)) made it clear that we were looking to implement *grouping sessions under a parent session* rather than straight up parallel sessions. Particularly for poster sessions, this is imo a better approach, rather than having users scroll through a confusing scroll-within-scroll interface. 
- The PR anchors poster sessions on a room-level configuration. This makes no sense to me at all. It introduces a new setting, which we always try to avoid. A new setting increases complexity in all dependent places that need to check for it!
- Using a scrollable list inside an already-scrolling schedule is, in my opinion, a bad way to display parallel sessions. It is surprising to users, does not survive filtering well, and makes printed versions of the schedule hard to read. 
- The PR contains completely unnecessary / stylistic changes like changing a random return statement.
- The PR contains code duplication and dead frontend code
- The PR only adheres to the project code style and structure at the surface level.
- The PR contains several tests that are essentially useless.

Sorry if this sounds harsh: it's not meant to be. pretalx is stable and carefully built software, so any change to how we treat some of our fundamental assumptions (e.g. "you can only have one session per time slot and room") need careful consideration and planning.

That is not to say that permitting parallel sessions might not *also* be a feature we could introduce at some point. It would be a completely separate feature from the implementation of poster sessions, though, and would almost certainly be implemented differently. For both features, I'm happy to discuss sponsored development or a less tool-assisted PR. If it's only down to prompting an LLM to do the work, though – I can do that, too :wink:

Issue #2758, opened 2026-09-09
Temporary failures in plugins handling e-mail sending lead to SMTP as fallback


badbadc0ffee:
### Problem description

A plugin that takes responsibility for sending out e-mails via `queuedmail_pre_send` (such as pretalx-RT) currently cannot signal back that it will send the message asynchronously or in particular that the delivery might be delayed due to a temporary failure.

`send_draft()` will always assume that it must send the mail via SMTP if the plugin doesn't set the sent timestamp or the state of the `QueuedMail` to `SENT`.

Plugins should also be able to signal that the have started sending the e-mail even if the transmission is not yet complete. `send_draft()` should honor the `SENDING` state and skip sending via SMTP in that case as well.

The plugin can then later set the state to `SENT` once it was successful or back to `DRAFT` if not, potentially using the `error_data` and `error_timestamp` to show what went wrong.

### Link/URL

[pretalx-RT](https://github.com/badbadc0ffee/pretalx-rt/compare/main...v2026.2.1-40c3) needs a way to show sporadic failures and _not_ fall back to SMTP in such cases.

### Browser version, operating system

FreeBSD

### Pretalx version

v2026.2.1

PR #2759, opened 2026-09-09
Fix sending of e-mails by plugin

badbadc0ffee:
Plugins can now signal that they're handling delivery by setting the mail to `SENDING` state, without needing mail.sent set immediately.

This avoids the mail also going out via SMTP if the plugin has a momentary problem or sends asynchronously.

This PR closes/references issue #2758.


badbadc0ffee:
rebased, added documentation and test coverage

rixx (approved):

Issue #2786, opened 2026-09-13
Availability dialogue missing help text and translation


badbadc0ffee:
### Problem description

The availability dialog is missing the help text and translation on the submitter profile page and in the organizer area.

The label and help text should be set similarly to those used for the room availability dialog.
The help test from the room dialog should be reused to avoid unnecessary new translations.

### Link/URL

https://pretalx.com/democon/me
https://pretalx.com/orga/event/democon/speakers/C0FFEE/

### Browser version, operating system

Firefox, Linux/BSD/Windows

### Pretalx version

2026.2.1

badbadc0ffee:
oops. has already been solved in 93075e5.

PR #2787, opened 2026-09-13
Add label and help text (incl. translations) to availabilities form

badbadc0ffee:
This PR closes/references issue #2786.

### DIFF
--- a/src/pretalx/common/forms/fields.py
+++ b/src/pretalx/common/forms/fields.py
@@ -421,6 +421,16 @@ class AvailabilitiesField(CharField):
         if "initial" not in kwargs and self.instance and self.event:
             kwargs["initial"] = self._serialize(self.event, self.instance)
 
+        kwargs.setdefault("label", _("Availability"))
+        kwargs.setdefault(
+            "help_text",
+            _(
+                "Please click and drag to mark your availability during the conference with "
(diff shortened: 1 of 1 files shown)


src/pretalx/common/forms/fields.py
```diff
@@ -496,21 +412,25 @@ class AvailabilitiesField(CharField):
...
         if "initial" not in kwargs and self.instance and self.event:
             kwargs["initial"] = self._serialize(self.event, self.instance)
 
+        kwargs.setdefault("label", _("Availability"))
+        kwargs.setdefault(
+            "help_text",
+            _(
+                "Please click and drag to mark your availability during the conference with "
+                "green blocks. We will try to schedule your slot during these times. You can "
+                "click a block twice to remove it."
```
rixx, line 430: This isn't how you remove blocks anymore. With the new UI, there is an explicit [x] button that gets shown on hover (or always on devices that do not support hover).

rixx (changes requested):
I think you're looking at outdated code for this one (v2026.2.1?)

The widget got its UI update in 93075e51c, which also includes an updated help text, which is also [set as default](https://github.com/pretalx/pretalx/blob/main/src/pretalx/common/forms/fields.py#L508) here on the form field.

